PR #1, opened 2018-10-21
import jest defaults from jest-config

SimenB:
I got a conflict when deleting this from Jest's source tree, so I think it makes sense to remove the hard-coding.

There are some comments at the top of the file about defaults, do you want more from it?


orta:
This seems 👍  to me

Issue #8, opened 2019-04-14
Addressing issue facebook/jest#4910


jackHedaya:
I'd like to reopen the issue found in https://github.com/facebook/jest/issues/4910 that discusses the use of string literals in test names.

seanpoulter:
Sure, any thoughts on how to move it forward? It has been a while since I looked at it and ended up with a bunch of questions in this comment:

https://github.com/facebook/jest/issues/4910#issuecomment-345478728.

connectdotz:
we have added template literal string parsing support a few months ago ([here](https://github.com/jest-community/jest-editor-support/blob/171f845655afe6929cd97a2bab7c3948e05e916f/src/parsers/babylon_parser.js#L58)), but just hasn't integrated the new package with vscode-jest, which I am trying to do now.

seanpoulter:
Oh heck yea! 🥇 Thanks @connectdotz!
Would you have capacity to review the PR @jackHedaya? 😉

jackHedaya:
Going to give it a shot now, though I recommend someone else double checking as this is my first Jest PR review.

seanpoulter:
How's it working for you now @jackHedaya?

jackHedaya:
Works fine now, thank you!

PR #9, opened 2019-04-17
refactoring prepare for vscode-jest integration

connectdotz:
## changes outline:
1. upgrade dependency (typescript, jest, babel, etc) to the latest versions.
1. merge `jest-test-typescript-parser` since it no longer exists in jest
    - why not create a separate `jest-test-typescript-parser` package like before?
      - its purpose is to support parsing functionality/format in `jest-editor-support`.
      - its logic resides in 1 file. 
      - babel-7 has a typescript plugin that could potentially replace this once we get around to test it.
1. expose a new `parse` function that can handle both js and ts file. 
1. supporting concurrent Runner ([vscode-jest/#442](https://github.com/jest-community/vscode-jest/issues/442))
1. fix test result parsing:
    - when no test is found, Runner omitted test result parsing that could cause clients waiting for test result indefinitely.
    - jest stderr message was truncated (arbitrarily?) prior to parsing that could lead to missing critical information such as test results output message.
1. <strike>enabled optional chaining (for developing `jest-editor-support` only, not test parsing)</strike>
1. incorporate jest location reporting. This can facilitate test matching process when test name match failed (such as template literal test: #8, [vscode-jest/#247](https://github.com/jest-community/vscode-jest/issues/247))
    - note: even though we can parse template literal test names, it doesn't mean we can find the matched assertion in the jest results! because the parser will report the raw string only while the test results contain the interpreted value. That is why we might still need to fall back to location-based matching mechanism...
1. add the ability to suppress Snapshot parsing error (often due to attempting to parse a typescript file as a snapshot), could still be turned on by a flag
1. added ci/coveralls/danger


connectdotz:
@orta or @seanpoulter please take a look, I finally got around to upgrade vscode-jest with the new `jest-editor-support` package, it's long overdue...

dangerfile.js
```diff
@@ -0,0 +1,24 @@
+import { danger, fail } from 'danger'
+import * as fs from 'fs'
+
+const pr = danger.github.pr
+const modified = danger.git.modified_files
+const bodyAndTitle = (pr.body + pr.title).toLowerCase()
+const trivialPR = bodyAndTitle.includes('#trivial')
+
+const codeOnly = (file ) => file.includes('.js')
```
stephtr: There is an unnecessary space. Even though it doesn't really matter, wouldn't `.endsWith` be better suited?

src/__tests__/parsers/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+jest.mock('../../parsers/babylon_parser', () => {
+  const mock_js_parse = jest.fn();
+  return { parse: mock_js_parse, mock_js_parse };
+})
+jest.mock('../../parsers/typescript_parser', () => {
+  const mock_ts_parse = jest.fn();
+  return { parse: mock_ts_parse, mock_ts_parse };
+})
+
+import {parse} from '../../parsers';
```
stephtr: Maybe one should somewhen lint the files, since the code style (spaces) is not always consistent.

src/__tests__/parsers/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+import { mock_js_parse } from '../../parsers/babylon_parser';
```
stephtr: It's also just an aesthetical thing, but do we have to call it "mock"? Can't we just import `parse`?

src/__tests__/parsers/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+describe('select parser', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  });
+  it('for .js or .jsx file', () => {
+    const files = ["abc.js", "abc.jsx"]
+    files.forEach(file => {
+      parse(file, undefined, true);
+      expect(mock_js_parse).toHaveBeenCalled();
```
stephtr: The mock should probably get reseted after each file.
By the way: Is there any reason for choosing double quotes for `files`?

src/parsers/typescript_parser.js
```diff
@@ -0,0 +1,138 @@
...
+  ParsedNode,
+  ParseResult,
+} from './parser_nodes';
+import type {ParsedNodeType} from './parser_nodes';
+
+export function parse(file: string, data: ?string): ParseResult {
+  const sourceFile = ts.createSourceFile(
+    file,
+    data ? data : readFileSync(file).toString(),
+    ts.ScriptTarget.ES3,
```
stephtr: Is there any reason for not using `ts.ScriptTarget.Latest`?

src/Runner.js
```diff
@@ -175,20 +175,21 @@ export default class Runner extends EventEmitter {
...
-    }
-    if (/^\s*Watch Usage\b/.test(str)) {
-      return messageTypes.watchUsage;
-    }
 
-    if (str.trim().startsWith('Test results written to')) {
-      return messageTypes.testResults;
-    }
-    return messageTypes.unknown;
+  findMessageType(buf: Buffer): MessageType {
```
glentakahashi, line 191: I think it's extremely unlikely, but technically this still might not catch all the files. Because the buffer can be flushed partway, it's possible that strings are actually split over multiple messages. Anecdotally, I've noticed the test output can be split mid-word. However, due to these messages usually being at the beginning/end of the test output, I don't think it'll be a problem.

Anyways, thanks for fixing this! I thought i was going insane trying to figure out why some of my projects would work but others wouldn't...

dangerfile.js
connectdotz: indeed, copied this file from `vscode-jest`, didn't pay too much attention to the syntax, good catch.

src/__tests__/parsers/index.test.js
connectdotz: good idea, I will add lint and lint-stage accordingly

src/__tests__/parsers/index.test.js
connectdotz: that was sloppy on my part, fixed.

src/__tests__/parsers/index.test.js
connectdotz: the clearAllMocks already happened in the outer describe's beforeEach()...

src/parsers/typescript_parser.js
connectdotz: I am not sure why it is ES3... this was just copied from the old repo, probably it hasn't caused any problem so nobody bothered to update it...? 

I am thinking it might be better to just leave it as it is to maintain backward compatibility; and besides, we might be moving away from this parser in favor of babel's typescript parser going forward...

src/Runner.js
connectdotz: you have a good point, I haven't seen these sentences split, but in theory, it could happen, especially if the output is large I guess... 

We could potentially buffer the sentence without EOL but that won't solve the problem when no-change message comes before the output-file-is-written message (now in a separate chuck)...

I am hesitated to add more code for a theoretical scenario that we haven't seen much evidence yet... let's make a note here and come back to it once there is more evidence.

glentakahashi: Yea, I was just making a comment here, I think the code you have is perfectly fine, just noting for future reference in case something ever happens here!

src/__tests__/parsers/index.test.js
stephtr: But in my opinion that is called only before each `it` block, not before each single `file`.

src/parsers/typescript_parser.js
stephtr: I think it doesn't matter at all since the code file doesn't get transpiled but we only obtain the AST.

src/__tests__/parsers/index.test.js
connectdotz: perhaps this helps: https://jestjs.io/docs/en/setup-teardown#scoping

connectdotz:
thanks for @stephtr and @glentakahashi reviewing this. 

@orta is there anything else you would like us address before merging?

src/__tests__/parsers/index.test.js
stephtr: But `beforeEach` nevertheless gets called in the following sample only once:

```js
describe('test', () => {
  beforeEach(() => {
    // executed only once
    console.log('beforeEach');
  });
  it('test', () => {
    [1, 2, 3].forEach(v => {
      // called three times
      console.log(v);
    });
  });
});
```

In the case of the test we are talking about that means, that it falsely succeeds in the case that:

* `mock_js_parse`, but not `mock_ts_parse` gets being called within parse for the first entry ("abc.js")
* and for the second entry ("abc.jsx") neither `mock_js_parse` nor `mock_ts_parse` gets called

since the mock functions only get reset before each `it` block, but not before each call of the `forEach` callback.
Or am I completely wrong?

connectdotz: ok, I am an idiot, I misunderstood, but now I see what you are talking about, you are absolutely right.

stephtr: It's fine, I also could have formulated it better in the first place.

connectdotz:
earth calling @orta ... 

there are many changes (see [vscode-jest/#442](https://github.com/jest-community/vscode-jest/issues/442)) pending on this PR, please let us know how do you want this PR to proceed...

stephtr:
Otherwise we could merge the PR due to the importance now (in a few days without response) and address possible changes later.

connectdotz:
ok, looks like @orta has abandoned us... maybe it's time to grow up... 

@SimenB @seanpoulter if you guys don't object, I will merge this PR and cut a new release `25.1.0` later this week.

src/Runner.js
```diff
@@ -40,7 +42,7 @@ export default class Runner extends EventEmitter {
     this._createProcess = (options && options.createProcess) || createProcess;
     this.options = options || {};
     this.workspace = workspace;
-    this.outputPath = tmpdir() + '/jest_runner.json';
+    this.outputPath = `${tmpdir()}/jest_runner_${this.workspace.outputFileSuffix || ''}.json`;
```
seanpoulter: Should we use a platform-agnostic path separator here?

src/Runner.js
```diff
@@ -168,27 +179,28 @@ export default class Runner extends EventEmitter {
     }
     if (process.platform === 'win32') {
       // Windows doesn't exit the process when it should.
-      spawn('taskkill', ['/pid', '' + this.debugprocess.pid, '/T', '/F']);
+      spawn('taskkill', ['/pid', `${this.debugprocess.pid}`, '/T', '/F']);
```
seanpoulter, line 183: Does the code formatter really suggest using a template literal `` `${arg}` `` over the explicit type coercion `String(arg)`? You ES-Next kids are so fancy! 🚀  https://eslint.org/docs/rules/no-implicit-coercion

src/Runner.js
```diff
@@ -168,27 +179,28 @@ export default class Runner extends EventEmitter {
...
+
+    const checks = [
+      [testResultsRegex, messageTypes.testResults],
+      [noTestRegex, messageTypes.noTests],
+      [watchUsageRegex, messageTypes.watchUsage],
+    ];
+
+    const str = buf.toString('utf8');
+    const check = checks.find(_check => _check[0].test(str));
+    return check ? check[1] : messageTypes.unknown;
```
seanpoulter: The leading underscore in `_check` and `[0]` in this line caught my eye:
```js
    const check = checks.find(_check => _check[0].test(str));
```

How do you feel about naming these?
```js
    const checks = [
      { regex: testResultsRegex, messageType: messageTypes.testResults },
      { regex: noTestRegex, messageType: messageTypes.noTests },
      { regex: watchUsageRegex, messageType: messageTypes.watchUsage },
    ];

    const str = buf.toString('utf8');
    const match = checks.find(({ regex })=> regex.test(str));
    return match ? match.messageType : messageTypes.unknown;
```

babel.config.js
```diff
@@ -2,4 +2,5 @@
 
 module.exports = {
   presets: ['@babel/preset-env', '@babel/flow'],
+  plugins: ["@babel/plugin-proposal-optional-chaining"],
```
seanpoulter: This makes me a little uncomfortable because TypeScript is slow to adopt "new" language features. Does it seem to be working out OK, or is that one of the reasons why we've disabled the validation in **.vscode/settings.json**?

src/Snapshot.js
```diff
@@ -84,41 +68,47 @@ const getArrayOfParents = path => {
...
+    }
     const state = {
       found: [],
     };
     const Visitors = {
-      Identifier(path, state, matchers) {
+      Identifier(path, _state, matchers) {
         if (matchers.indexOf(path.node.name) >= 0) {
-          state.found.push({
+          _state.found.push({
```
seanpoulter, line 111: Is it just me, or does this object assignment with the `Identifier` function look a bit off? It took a little bit to realize it's a function declaration, and sort out the `_state` vs `state`. With that in mind, how about we clean up this painful block a bit:

Here's what we've got:
```js
    const Visitors = {
      Identifier(path, _state, matchers) {
        if (matchers.indexOf(path.node.name) >= 0) {
          _state.found.push({
            node: path.node,
            parents: getArrayOfParents(path),
          });
        }
      },
    };

    traverse(fileNode, {
      enter: path => {
        const visitor = Visitors[path.node.type];
        if (visitor != null) {
          visitor(path, state, this._matchers);
        }
      },
    });
```

It looks like we're trying to add a node and parents to the `state.found` array if:
* The node is an identifier (`node.path.type === 'Identifier'`)
* It is one of the built-in or custom matchers in the `this._matchers` list

If we were to take these two criteria into a private class method:
```js
  _isSnapshotMatcher(path) {
    return path && path.node && path.node.type === 'Identifier'
      && this._matchers.indexOf(path.node.name) >= 0;
  }
```
then we can reduce that whole block to:
```js
    traverse(fileNode, {
      enter: path => {
        if (this._isSnapshotMatcher(path)) {
          state.found.push({
            node: path.node,
            parents: getArrayOfParents(path),
          });
        }
      },
    });
```

<br/>

If you really wanted to go overboard, why do we need the `.` in `state.found` if if doesn't look like we're using any other members on the `state` object?

src/__tests__/parsers/index.test.js
```diff
@@ -0,0 +1,61 @@
...
+        expect(js_parse).toHaveBeenCalled();
+        expect(ts_parse).not.toHaveBeenCalled();
+        jest.clearAllMocks();
+      });
+    });
+    it('throw exception in strict mode', () => {
+      const files = ['abc', 'abc.ttsx'];
+      files.forEach(file => {
+        expect(() => parse(file, undefined, true)).toThrow();
+        jest.clearAllMocks();
```
seanpoulter, line 57: It's likely not a problem in this case, but I usually suggest we keep the test teardown (`jest.clearAllMocks();`) out of the test just in case something fails during the execution.

seanpoulter:
Wow, this looks like you've put in a lot of work @connectdotz. Bravo! 👏 

Sorry I haven't had the bandwidth to review so far, and will likely be even more quiet in the months ahead. I've tried having a look tonight, but given the size I'm not making a lot of progress. It seems like there's a sizeable amount of formatting changes.

<br/>

> @SimenB @seanpoulter if you guys don't object, I will merge this PR and cut a new release 25.1.0 later this week.

Personally, I feel a bit "erm... hold on" about it. @stephtr and @glentakahashi haven't approved the review, and I've maybe made it 1/4 of the way through. I'm pretty sure you've done a good job, but I don't know what the risk is to VS Code Jest or Majestic if we get something wrong. Since I don't know... "erm... hold on." Do you think someone from the Raathigesh/majestic team could help @Raathigesh, or do you have any interns we can buy some pizza @SimenB?

SimenB:
I'm not in a position to review this, sorry.

I'm a big fan of just releasing and fixing later. Can release as `next` or even a new major if you're afraid of breaking things?

---

Might be a good idea to land the formatting and ci changes separately, so one PR is focused just on that, and one on actual code changes. Makes it easier to review since the noise is greatly reduced. Also useful when looking through history later for the same reasoning

Raathigesh:
Hey,

Majestic doesn't use jest-editor-support anymore so I'm afraid I might not be really helpful with this.

seanpoulter:
OK. Thanks for the quick reply @Raathigesh (and congrats on Atlassian if that's new).

connectdotz:
nice to see many of you here 😄 , just when we started to feel lonely...

Yeah it was stupid of me to add the internal improvement in the same PR, but I wanted to be sure that I didn't break anything so introduced ci; the inconsistency of formatting and coding style have bugged me for a while and also got pointed out during the review thus decided to finally pull the trigger and added more fixes... The thoughts of delaying definitely cross my mind but thinking this is precisely why we haven't done it, it will be a pain to review but might as well be this time since we will be doing lots of testing for these changes anyway... Hopefully, the following can bring some comfort to ease the review pain:
- I have been running a beta vscode-jest plugin ([vscode-jest/#447](https://github.com/jest-community/vscode-jest/pull/447)) with this PR for over a month exclusively as my daily development tool, they worked smoothly. Granted it is only on mac and limited exposure (however it did cover single and monorepo use cases), but I am relatively confident that it didn't break the common use cases. 
- I am planning to ask for more testers once we cut the next `vscode-jest` release so we will have more chances to address issues if any.
- most changes you saw are indeed introduced by prettier/eslint, and I did let them perform auto fix for all the src/test due to the amount of violation. I would admit that I did not revisit the changes done by these tools, @seanpoulter comments provided a good opportunity to discuss it and I will address them inline.

@seanpoulter I am more than happy to wait for your review, this PR has been sitting idle for almost a month **begging** for reviews. I know PR is not supposed to be merged by its author, thus keep pinging you guys... knowing many changes on vscode-jest is pending on this PR and nobody seemed to want to take responsibility to move forward, out of desperation, I decided to take @stephtr advice to merge and unblock... seems like you have some concern about not seeing explicit approval before proceeding with merge, I agree with you wholeheartedly! I shall not perform the merge and let you and @stephtr decide how to proceed.

src/Runner.js
connectdotz: good point, I wonder why it didn't break on windows before? oh well, we should do the right thing regardless...

src/Runner.js
connectdotz: this is most likely determined by the eslint, in `.eslintrc.js`:
```
extends: ['airbnb-base', 'plugin:flowtype/recommended', 'prettier', 'plugin:prettier/recommended'],
```
I don't have any preference actually, it seems fine to me either way as long as we are consistent in the code base.

src/Runner.js
connectdotz: good, done.

babel.config.js
connectdotz: I missed using optional-chaining from Swift, but no need to get too far ahead of ourselves here... done

src/Snapshot.js
connectdotz: I definitely agree with you this code can be made more readable, but it has been like this for a while, all I did is to rename `state` to `_state` to conform to the lint rule "no-shadow"...  are you sure we want to do the refactoring here, given the PR is already large and this is not related to what we try to fix?

src/__tests__/parsers/index.test.js
connectdotz: agree but you can see why we had to make an exception here.

src/Snapshot.js
seanpoulter: Hah! Thanks for calling me on the scope creep. Your change is AOK. 👌
I'll open a new PR .. someday. 😜

connectdotz:
> but I don't know what the risk is to VS Code Jest or Majestic if we get something wrong. Since I don't know... "erm... hold on."

I think these have been answered by @Raathigesh and myself in the comments earlier. Of course, the only sure way to find out is to release and test in the real world.

>  I shall not perform the merge and let you and @stephtr decide how to proceed.

hmmm... should I add procrastination is not one of them... 😏 

I know when changes are big we, as reviewers, all worried that our reviews might be incomplete and we might be missing something thus hope the others can decide. Giving we are the collaborators/owners of these packages, we need to face our fear heads on so we can move forward... Every day we sit around without any action is the day it could have been tested by many others out there...

connectdotz:
@seanpoulter looks like you are busy... 

(I feel this is the time to test our open source resolve and wisdom 😓 )

If nobody has time to review further, then I agree at this point the best approach is probably "release now and fix later" as @stephtr and @SimenB have suggested. We can go with `26.0.0-beta` just to be safe. 

For all the people who care about `jest-editor-support` and `vscode-jest`, please do speak up if you have better suggestions... If not, I will try to merge and release it later this weekend.

orta:
I'm always a strong believer in shipping things when you feel good about it, and prioritizing that - if it breaks the occasional build, that's an acceptable trade off.  Makes it easier to ship often and smaller releases.

connectdotz:
ok code is merged and [release](https://github.com/jest-community/jest-editor-support/releases) is created, but I don't have permission to publish to npm, @orta can you help?

connectdotz:
ok, [jest-editor-support@26.0.0-beta](https://www.npmjs.com/package/jest-editor-support/v/26.0.0-beta) is published to npm.

Issue #11, opened 2019-07-09
License of some modules


luisapace:
Hi, my team would use jest-editor support and to approve that I've to verify all its licenses.

Now I've seen that some modules (i.e.  jest-editor-support/src/__tests__/snapshot.test.js) show this header:
 **_* This source code is licensed under the BSD-style license found in the
 * LICENSE file in the root directory of this source tree. An additional grant
 * of patent rights can be found in the PATENTS file in the same directory._**

other modules instead refers to MIT license, but in the root directory of the source I cannot see any license file.

Could you please clarify me what are the licenses? 

Thanks for your support on this.

connectdotz:
hmmm... I saw some files still have facebook copyright, which is understandable since it was part of [jest](https://github.com/facebook/jest). But now it has been moved out to `jest-community`, should we update all the copyrights accordingly? @orta, what is the overall license policy for `jest-community`? I saw some repo using MIT, does that apply to this repo as well?

orta:
I'm gonna cc in @cpojer for this one, FB copyrights are not something I'm going to take a guess on

cpojer:
Change it to MIT.

orta:
Fixed with #15

luisapace:
Many thanks !

luisapace:
Many thanks !

PR #15, opened 2019-07-22
Create LICENSE

connectdotz:
addressing #11

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Jest Community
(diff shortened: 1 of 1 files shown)


orta:
Thanks, luckily this repo is already MIT - https://github.com/jest-community/jest-editor-support/blob/master/package.json#L8

The goal in #11 is to effectively remove the FB copyright headers on those files so that they don't have competing licenses 👍

orta:
Actually - on a second look - I think you're right, let's do this 💯

PR #18, opened 2019-09-23
Add custom short message for @testing-library/react

jmarceli:
I think that short messages for @testing-library/react should have more content than the others. 

The current implementation looks as follows:

no-testing-library-support

While after this MR it will look like that:

testing-library-support

Together with https://github.com/jest-community/vscode-jest/pull/501 it should provide a complete solution for https://github.com/jest-community/vscode-jest/issues/493

Please let me know how do you feel about this change.

### DIFF
--- a/src/test_reconciler.js
+++ b/src/test_reconciler.js
@@ -105,6 +105,11 @@ export default class TestReconciler {
       return 'New snapshot is ready to write';
     }
 
+    // Improve @testing-library/react error messages handling
+    if (string.startsWith('Error: Unable to find an element with the text')) {
+      return string;
+    }
+
     return string
       .split('\n')
       .splice(2)
--- a/src/test_reconciler.js
+++ b/src/test_reconciler.js
@@ -13,6 +13,10 @@ import type {TestFileAssertionStatus, TestAssertionStatus, TestReconciliationSta
 
 import type {FormattedAssertionResult, FormattedTestResults} from '../types/TestResult';
 
+// Regex which defines which short messages are considered to be valid
+// \s\S - stands for any character including new line https://stackoverflow.com/questions/1068280/javascript-regex-multiline-flag-doesnt-work
+const validShortMessage = /expected[\s\S]*received/i;
+
 /**
  *  You have a Jest test runner watching for changes, and you have
  *  an extension that wants to know where to show errors after file parsing.
@@ -105,8 +109,7 @@ export default class TestReconciler {
       return 'New snapshot is ready to write';
     }
 
-    // Standard Jest error message
-    if (string.startsWith('Error: expect')) {
+    if (validShortMessage.test(string)) {
       return string
         .split('\n')
         .splice(2)
--- a/src/test_reconciler.js
+++ b/src/test_reconciler.js
@@ -14,8 +14,7 @@ import type {TestFileAssertionStatus, TestAssertionStatus, TestReconciliationSta
 import type {FormattedAssertionResult, FormattedTestResults} from '../types/TestResult';
 
 // Regex which defines which short messages are considered to be valid
-// \s\S - stands for any character including new line https://stackoverflow.com/questions/1068280/javascript-regex-multiline-flag-doesnt-work
-const validShortMessage = /expected[\s\S]*received/i;
+const validShortMessage = /Expected.*[R|r]eceived/s;
 
 /**
  *  You have a Jest test runner watching for changes, and you have
--- a/src/test_reconciler.js
+++ b/src/test_reconciler.js
@@ -105,18 +105,19 @@ export default class TestReconciler {
       return 'New snapshot is ready to write';
     }
 
-    // Improve @testing-library/react error messages handling
-    if (string.startsWith('Error: Unable to find an element with the text')) {
-      return string;
+    // Standard Jest error message
+    if (string.startsWith('Error: expect')) {
+      return string
+        .split('\n')
+        .splice(2)
+        .join('')
+        .replace(/\s\s+/g, ' ')
+        .replace('Received:', ', Received:')
+        .split('Difference:')[0];
     }
 
-    return string
-      .split('\n')
-      .splice(2)
-      .join('')
-      .replace(/\s\s+/g, ' ')
-      .replace('Received:', ', Received:')
-      .split('Difference:')[0];
+    // Non standard message
+    return string;
   }
 
   // Pull the line out from the stack trace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Bug-fixes within the same version aren't needed
   projects that link with this package no longer need to add `jest-test-typescript-parser` package separately. The newly exposed `parse` function can select the proper parser based on file extension.
 
 * fix a few race condition errors in Runner (#9) - connectdotz
-  - concurrent Runner output to the same hard coded output file: added optional 'outputFileSuffix' parameter. 
+  - concurrent Runner output to the same hard coded output file: added optional 'outputFileSuffix' parameter.
   - jest process output parser sometimes failed to identify testResults message.
 
 * `TestReconciler` will now report test locations from jest '--testLocationInResults' output
@@ -26,6 +26,8 @@ Bug-fixes within the same version aren't needed
 
   `getSettings` now simply returns a promise resolving to jest's config.
 
+* custom short message for @testing-library/react - jmarceli
+
 -->
 
 ### 25.0.0
@@ -37,8 +39,7 @@ in your head, consider this a 1.0 kinda thing.
 
   This is because the codebase was migrated with Jest, and now uses functions
   from the master builds that don't seem to be available on the latest production
-  versions of jest-snapshot/. If someone wants to backport this to the prod builds 
+  versions of jest-snapshot/. If someone wants to backport this to the prod builds
   look at the usage of the `buildSnapshotResolver`
 
 - Adds the ability to parse describe blocks - https://github.com/facebook/jest/pull/7215
-
--- a/babel.config.js
+++ b/babel.config.js
@@ -6,4 +6,5 @@ module.exports = {
     ],
     '@babel/flow',
   ],
+  plugins: ['@babel/plugin-transform-dotall-regex']
 };
--- a/fixtures/failing-jsons/failing_testing-library-react.json
+++ b/fixtures/failing-jsons/failing_testing-library-react.json
@@ -1,0 +1,54 @@
+{
+  "numFailedTestSuites": 1,
+  "numFailedTests": 1,
+  "numPassedTestSuites": 0,
+  "numPassedTests": 0,
+  "numPendingTestSuites": 0,
+  "numPendingTests": 0,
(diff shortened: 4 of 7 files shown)


src/test_reconciler.js
```diff
@@ -105,6 +105,11 @@ export default class TestReconciler {
       return 'New snapshot is ready to write';
     }
 
+    // Improve @testing-library/react error messages handling
+    if (string.startsWith('Error: Unable to find an element with the text')) {
+      return string;
+    }
+
```
seanpoulter, line 120: Interesting idea @jmarceli 👏 . It seems like we're missing an extension point either deciding what to keep or deciding what to show. I've been out of the fold for a while. Perhaps @connectdotz or @stephtr have a strong opinion about our architectural direction to address this.

jmarceli: I'm aware that such message based decisions are not future proof but I didn't find any other way to address library specific tests and current short message is not very user friendly. If I'm missing something just let me know.

connectdotz: hmmm... I am not sure about adding this non-jest pattern check here, however, I do see a general assumption that maybe should be reexamined...  

It appeared that this class considered only the jest specific format, evidence in the explicit patterns it used to extract short and terse messages. From your use case, it is apparent that indeed there could be custom matcher or 3rd party tools that could provide different error messages. Instead of selectively passing these message by looking for their patterns, I wonder would it be better to just pass through any non-jest messages because there is no way we could have guessed a good "short" message from the unknown... 

It seems we might be able to reasonably guess if a message is given by jest by examing some common patterns like we did [here](https://github.com/jest-community/jest-editor-support/blob/591355c4ef81055cce30f25ea844914b2c5b72da/src/test_reconciler.js#L81), when a message doesn't pass the "jest check", no need to continue parsing, we could just set short/terse/message to be the message itself...

would that work?

jmarceli: It seems fine for me and it will work in my case but I'm not sure how would you like to detect standard Jest message format?
In your link the line `message.split('   at', 1)[0].trim()` is highlighted but I have no idea how it can be used for standard Jest messages detection (as far as I can tell @testing-library messages also has `'   at'` part).
Based on examples provided in the other tests I've decided to go with a different way of detecting standard Jest messages.
@connectdotz please let me know what do you think about that and if it's required how would you like to modify that.

connectdotz: > but I'm not sure how would you like to detect standard Jest message format?

that's a good question, you are right that it is hard to say definitively... Maybe let's tackle it from another angle. 

Looking at the original algorithm, I think the intent of this short message is to show the "Expected vs Received" information and by-passing stack trace. So to keep this goal in mind, we can ask ourselves if we can produce this "target" message format, if not, then we are better off to just use the full message...

what do you think?

jmarceli: I came up with this not-so-clever regex `/expected[\s\S]*received/i` please let me know if you have something more sophisticated in your mind. For the current test suite, it is working fine.

src/test_reconciler.js
```diff
@@ -13,6 +13,10 @@ import type {TestFileAssertionStatus, TestAssertionStatus, TestReconciliationSta
 
 import type {FormattedAssertionResult, FormattedTestResults} from '../types/TestResult';
 
+// Regex which defines which short messages are considered to be valid
+// \s\S - stands for any character including new line https://stackoverflow.com/questions/1068280/javascript-regex-multiline-flag-doesnt-work
+const validShortMessage = /expected[\s\S]*received/i;
```
stephtr: If we would stick to this, it's maybe worth adding [@babel/plugin-transform-dotall-regex](https://babeljs.io/docs/en/babel-plugin-transform-dotall-regex) and just use the [dotAll flag](https://github.com/tc39/proposal-regexp-dotall-flag) (/s).

src/test_reconciler.js
stephtr: Even though I like this approach, I see the issue of false positives as soon as a custom string contains the two words. We probably won't get it to work perfectly (which isn't necessary), but I would at least precise it to `/Expected: .*\nReceived: /s` (and hope, that Jest is never going to localize their messages 😉).

jmarceli: I'm not sure about `/Expected: .*\nReceived: /s` as it won't work for current test suite. I mean this message `Expected value to be falsy, instead received\n  2`. The only improvement I can make would be `/Expected .*[r|R]eceived/s` but I don't think it is an improvement worth adding.

src/test_reconciler.js
jmarceli: Are you sure about adding another plugin just for this regex? I would stay with the current approach as it is working well without adding new dependencies.

src/test_reconciler.js
stephtr: You are completely correct, I overlooked that.

src/test_reconciler.js
stephtr: I see it less as adding another plugin but as using ES2018 in order to simplify the syntax. But that is just my personal point of view.

jmarceli: That's true, the syntax is indeed a bit simpler with dotAll flag. I've added this improvement as well as removed case-insensitivity according to the previous suggestion.

stephtr: 👍
> The only improvement I can make would be /Expected .*[r|R]eceived/s but I don't think it is an improvement worth adding.

That was true, but I'm fine with both version.

connectdotz:
hi guys,

I ran with this PR for a little bit over the weekend, and I must say that I am actually much preferred to see the unshortened message (minus stacktrace)  than the shorten one, almost on all occasions! Granted this might be just personal preference, but here are my reasons:

Let's just look at these examples:
- a best-case scenario:
  ```
  it('dummy 2', () => {
    expect(true).toBe(false);
  });
  ```
  our code produced simple and precise message, its the text book use-case:
  ```
  Expected: false, Received: true
  ```
  however, if we don't shorten and this is what we will get:
  ```
  expect(received).toBe(expected) // Object.is equality

  Expected: false
  Received: true
  ```
  Yes, this is a bit verbose but hardly less clear, on the contrary, I think it is more clear and provided more info.

- when we don't shorten, how bad is it? another example:
  ```
  it('dummy', () => {
    expect(() => {}).resolves.toBe(false);
  });
  ```
  an unshortened message will be 
  ```
  expect(received).resolves.toBe()

  Matcher error: received value must be a promise

  Received has type:  function
  Received has value: [Function anonymous]
  ```
  is it bad? IMHO, not at all: in vscode, the message is also nicely presented in the popup, it's pretty easy to read.

- does shorten messages sometimes hurt more than help? 
  ```
  it('dummy 3', () => {
    expect([1, 2, 3]).toBe([1, 2]);
  });
  ```
  the shortened message:
  ```
  - Expected+ Received Array [ 1, 2,+ 3, ]
   ```
  it's a bit encrypted, however, experienced developers could probably still able to figure out with a little bit of effort. But wait, let's compare with the unshortened message:
  ```
    expect(received).toBe(expected) // Object.is equality

    - Expected
    + Received

      Array [
        1,
        2,
    +   3,
      ]
  ```
  I found the later is a lot more readable! 😄 

  I also have seen sometimes the shorten message algorithm merged lines without any white-space which made it pretty confusing, to say the least.
----
Thanks for @jmarceli and @seanpoulter bring this to our attention, I have a total change of heart on the short messages. I am now thinking we can just return the message as it is (after truncated stack-trace). This will provide more useful information to our users while simplifying the code and improving performance!

We can always come back with more optimization, if needed, such as limiting # of lines, removing empty line etc, but I am not convinced that we need to do this right now...

What do you guys think? maybe we should all running with the new "unshortened message" for a few days to ensure there is no obvious issue and if we really like it better.

stephtr:
Apparently the original idea was ["everything we can to try make a one-liner from the error report"](https://github.com/orta/jest/blob/873e023ab896a4cc7bfb03fa8733dee9c3d9007e/packages/jest-editor-support/src/TestReconciler.js#L99).
For VS Code I would be fine with showing the unshortened messages, especially since we would get rid of some quite specific code in return. My only concern would be, that there is some other program out there relying on jest-editor-support and the approach of having a one-liner as a message.

connectdotz:
@stephtr has a good point, I went back to `vscode-jest` and looks like the problem is more in `vscode-jest` than here... 

the fact that we even have the "terse" vs "short" I guess is precisely for inline vs. hover (popup) message display.  the inline is a comment-style one-liner (terseMessage) while hovering/popup message can be multi-lin and should have been the shortMessage but is the same terseMessage right now. (see [here](https://github.com/jest-community/vscode-jest/blob/5c786d21407a8fc948cd9f5617cbdf92d2acdd71/src/JestExt.ts#L374), [here](https://github.com/jest-community/vscode-jest/blob/5c786d21407a8fc948cd9f5617cbdf92d2acdd71/src/diagnostics.ts#L54), maybe more...)

I think it is pretty hard to extract a multi-line error message into a good meaningful single liner... this PR is a testimony. The good news is I don't think we need to address this hard problem once we fix `vscode-jest`. Users will get a lot more detailed information when hovering over the error, so the terseMessage is sort of like "best-effort" and one can always get the actual error message in the hovering popup, I think it's an acceptable user-experience, at least no worse than what we have now.

In this case, I think we can scratch this PR and open a new PR in `vscode-jest` to address the hovering message bug, @jmarceli what do you think?

jmarceli:
In my opinion, showing multiline message on hover in VS code is definitely the way to go. I agree that those one-liners will never be as readable as their multiline counterparts espacially in case of "advanced" plugins like @testing-library/react.

The only question which in my opinion remains unanswered is whether we want to merge my MR or leave it as it is :) According to the @stephtr comment I would be happy to remove terse message altogether and simplify codebase but still there might be some other tools which depend on that feature. Maybe this is a good point for a new major release?

I will feel no regret if you leave my PR without merging, but maybe it is worth adding to the current version as it improves one-liners a bit and release the next major version without terse messages support.

connectdotz:
@jmarceli  I think we all agree that terseMessage is not ideal right now and, if possible, could be phased out eventually... therefore, we probably should just leave it alone and close this PR for now.

I know you must have spent quite some time working on this PR, our apology for not realizing the correct path forward sooner. But your effort definitely is not wasted! Without this PR we would not have discovered the hovering message bug! If you are not too frustrated, could we work with you to finish what this PR started in `vscode-jest` instead?

jmarceli:
No problem. I'll try to make a PR to `vscode-jest` which will use regular message instead of its terse version. Maybe I'll find some spare time in the following week.
What do you think about disabling the inline error message preview? (this red text next to the tested line) Thanks to that `vscode-jest` will not use terse message version at all.

connectdotz:
> What do you think about disabling the inline error message preview? (this red text next to the tested line)

personally I found it annoying and usually turned it off, but that could be just me...However, if you guys all feel the same, we could try to set the `jest.enableInlineErrorMessages` default to false (in package.json) in the next major release and see how it goes...

jmarceli:
To be honest I didn't notice that this is optional. I think that we could try to disable this option by default and see what happens. Personally I don't see much value in those inline messages.

PR #19, opened 2019-10-06
Update babel, add core-js as explicit dependency

stephtr:
This PR updates babel and additionally includes `core-js` as an explicit dependency, which is necessary when using `useBuiltIns`.


connectdotz (approved):
seems reasonable.

PR #20, opened 2019-10-06
Bump `set-value` version

stephtr:
According to [CVE-2019-10747](https://nvd.nist.gov/vuln/detail/CVE-2019-10747) `set-value@<2.0.1` is vulnerable and should be updated, which is done by this PR.


seanpoulter (approved):
Their docs say to bump to v3, but 2.0.1 [seems OK](https://www.npmjs.com/advisories/1012/versions).

stephtr:
Thanks for checking that. The updated dependents still need v2, so we can't update to v3 anyhow.

PR #21, opened 2019-10-12
fixed handling of quoted commands with spaces

### LINKED ISSUE #342: vscode-jest does not start when the workspace folder name (multi-root) contains blanks
### Environment

1. `node -v`: 10.5
2. `npm -v`: 6.1
3. VSCode: 1.24.1
4. Operating system: Windows 10 1709

### LINKED ISSUE #360: vscode-jest stopping at launch of VSCode
### Environment

1. `node -v`: v9.4.0 
1. `npm -v`: 6.3.0
1. VSCode: 1.25.1
1. Operating system: Windows 7 x64

### Prerequisite
- are you able to run jest test from command line? yes
- how do yo run your tests from command line? yarn test

###  Steps to Reproduce
1. Install Jest as dev-dependency
2. Create jest.config.js with following:
`module.exports = {
    verbose: true,
    testURL: "http://localhost/",
} `

### Expected Behavior
vscode-jest should work 

### Actual Behavior
After opening project with jest in VSCode:
 - vscode-jest stopped and printed the following: `Starting Jest in Watch mode failed too many times and has been stopped.`
- after reloading VSCode 100+ times - vscode-jest started working properly
- after another VSCode window reload vscode-jest stopped again

### LINKED ISSUE #426: Plugin won't run on Windows when project is in a path with a space in it.
### Environment

1. `node -v`: [v8.11.3]
1. `npm -v`: [v6.8.0]
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): [n/a]

1. Operating system: [Windows 8.1]

### Prerequisite
- are you able to run jest test from command line? [yes]
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) [npm run test]

### Steps to Reproduce

1. Place your project in a folder containing a space, eg: e:\My Websites\myproject.
Plugin attempts to run multiple times before giving up, saying it can't find 'e:\My'.

### Expected Behavior

Plugin should be able to handle Windows paths

### Actual Behavior

Complains it can't find 'e:\My'

--- 

The fastest (and the most fun) way to resolve the issue is to submit a pull-request yourself. If you are interested, feel free to check out the [contribution guide](https://github.com/jest-community/vscode-jest/CONTRIBUTING.md), we look forward to seeing your PR...

### LINKED ISSUE #489: Jest not working because of spaces in project path
### Environment

1. `node -v`: v8.15.1
2. `npm -v`: 5.6.0
3. `npm ls jest`:
```
C:\Users\Pardhu Chowdary
--` (empty)
```
4. your vscode-jest settings if customized:
 `"jest.pathToJest": "npm run test --"`

5. Operating system: Windows 10

### Prerequisite
- are you able to run jest test from command line?
No. When I run just `jest` it gives `bash: jest: command not found`. I am using git bash btw.
- how do you run your tests from command line? 
`npm run test`
Although I'm not able to run `jest` from command line, `npm run test` works fine (detects and runs all the tests)

### Steps to Reproduce

1. Installed Jest extension via marketplace.
2. Added `"jest.pathToJest": "npm run test --"` to settings.json.
3. Output in the output panel (ctrl+shift+u) when it runs:
>&gt; expense-manager@1.0.0 test e:\Dev\expense-manager
>&gt; jest --coverage --verbose "--testLocationInResults" "--json" "--useStderr" "--outputFile" "C:\Users\Pardhu" "Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json" "--no-color"
>
>
>No tests found, exiting with code 1
>Run with `--passWithNoTests` to exit with code 0
>In E:\Dev\expense-manager
>  46 files checked.
>  testMatch:  - 0 matches
>  testPathIgnorePatterns: \\node_modules\\ - 46 matches
>  testRegex: (\\\_\_tests\_\_\\.*|(\.|\\)(test|spec))\.ts$ - 1 match
>Pattern: Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json - 0 matches

I'm pretty sure this is caused due to the space in the path: 
`C:\Users\Pardhu Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json`
Tried the fix as suggested in #342 
### Relevant Debug Info

Partial package.json:
```
  "main": "dist/src",
  "scripts": {
    "prebuild": "tslint -c tslint.json -p tsconfig.json --fix",
    "build": "tsc",
    "prestart": "npm run build",
    "start": "node .",
    "test": "jest --coverage --verbose",
    "dev": "tsc-watch --onSuccess \"node ./dist/index.js\""
  },
  "dependencies": {
    "body-parser": "^1.19.0",
    "colors": "^1.3.3",
    "cookie-parser": "^1.4.4",
    "express": "^4.17.1",
    "firebase-admin": "^8.3.0",
    "mongoose": "^5.6.0"
  },
  "devDependencies": {
    "@types/cookie-parser": "^1.4.1",
    "@types/express": "^4.17.0",
    "@types/jest": "^24.0.15",
    "@types/mongoose": "^5.5.11",
    "@types/node": "^12.6.8",
    "jest": "^24.8.0",
    "nyc": "^14.1.1",
    "ts-jest": "^24.0.2",
    "ts-node": "^8.3.0",
    "tslint": "^5.18.0",
    "typescript": "^3.5.3"
  },
  "jest": {
    "verbose": true,
    "transform": {
      "^.+\\.tsx?$": "ts-jest"
    },
    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.ts$",
    "moduleFileExtensions": [
      "ts",
      "tsx",
      "js",
      "jsx",
      "json",
      "node"
    ]
  }
```

### Expected Behavior
Tests should appear in the sidebar

### Actual Behavior
Nothing appears. Status bar gets stuck with `Jest: -\|/ Starting watch mode`

--- 

The fastest (and the most fun) way to resolve the issue is to submit a pull-request yourself. If you are interested, feel free to check out the [contribution guide](https://github.com/jest-community/vscode-jest/CONTRIBUTING.md), we look forward to seeing your PR...

omjadas:
fixes jest-community/vscode-jest#342
fixes jest-community/vscode-jest#360
fixes jest-community/vscode-jest#426
fixes jest-community/vscode-jest#489
#trivial

### DIFF
--- a/src/Process.js
+++ b/src/Process.js
@@ -23,32 +23,16 @@ export const createProcess = (
   args: Array<string>,
   options?: SpawnOptions = {}
 ): ChildProcess => {
-  // A command could look like `npm run test`, which we cannot use as a command
-  // as they can only be the first command, so take out the command, and add
-  // any other bits into the args
-  const runtimeExecutable = workspace.pathToJest;
+  let runtimeExecutable = workspace.pathToJest;
 
-  // "[^"\\]*                 Matches a quote followed by zero or more
-  //                          characters that are not quotes or backslashes
-  // (?:\\[\S\s][^"\\]*)*"    Matches a backslash followed by any character,
-  //                          followed by zero or more characters that are not
-  //                          quotes or backslashes ('[\S\s]' is used instead of
-  //                          '.' so that newlines are also matched). This
-  //                          ensures that escaped quotes are handled correctly
-  // (?:\\\s(?!\s+)|\S)       Matches a backslash followed by a single
-  //                          whitespace character, or a non-whitespace
-  //                          character
-  const parameters = runtimeExecutable.match(
-    /(?:"[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|(?:\\\s(?!\s)|\S))+/g
-  ) || [''];
-  const command = parameters[0];
-  const initialArgs = parameters.slice(1);
-  const runtimeArgs = [].concat(initialArgs, args);
+  if (args.length) {
+    runtimeExecutable += ` ${args.join(' ')}`;
+  }
 
   // If a path to configuration file was defined, push it to runtimeArgs
   if (workspace.pathToConfig) {
-    runtimeArgs.push('--config');
-    runtimeArgs.push(workspace.pathToConfig);
+    runtimeExecutable += ' --config ';
+    runtimeExecutable += workspace.pathToConfig;
   }
 
   // To use our own commands in create-react, we need to tell the command that
@@ -59,13 +43,13 @@ export const createProcess = (
   const spawnOptions = {
     cwd: workspace.rootPath,
     env,
-    shell: options.shell,
+    shell: true,
   };
 
   if (workspace.debug) {
     // eslint-disable-next-line no-console
-    console.log(`spawning process with command=${command}, args=${runtimeArgs.toString()}`);
+    console.log(`spawning process with command=${runtimeExecutable}`);
   }
 
-  return spawn(command, runtimeArgs, spawnOptions);
+  return spawn(runtimeExecutable, [], spawnOptions);
 };
--- a/src/Process.js
+++ b/src/Process.js
@@ -18,16 +18,12 @@ import ProjectWorkspace from './project_workspace';
  */
 // eslint-disable-next-line import/prefer-default-export
 export const createProcess = (workspace: ProjectWorkspace, args: Array<string>): ChildProcess => {
-  let runtimeExecutable = workspace.pathToJest;
-
-  if (args.length) {
-    runtimeExecutable += ` ${args.join(' ')}`;
-  }
+  const runtimeExecutable = [workspace.pathToJest, ...args];
 
   // If a path to configuration file was defined, push it to runtimeArgs
   if (workspace.pathToConfig) {
-    runtimeExecutable += ' --config ';
-    runtimeExecutable += workspace.pathToConfig;
+    runtimeExecutable.push('--config');
+    runtimeExecutable.push(workspace.pathToConfig);
   }
 
   // To use our own commands in create-react, we need to tell the command that
@@ -43,8 +39,8 @@ export const createProcess = (workspace: ProjectWorkspace, args: Array<string>):
 
   if (workspace.debug) {
     // eslint-disable-next-line no-console
-    console.log(`spawning process with command=${runtimeExecutable}`);
+    console.log(`spawning process with command=${runtimeExecutable.join(' ')}`);
   }
 
-  return spawn(runtimeExecutable, [], spawnOptions);
+  return spawn(runtimeExecutable.join(' '), [], spawnOptions);
 };
--- a/src/Process.js
+++ b/src/Process.js
@@ -27,8 +27,19 @@ export const createProcess = (
   // as they can only be the first command, so take out the command, and add
   // any other bits into the args
   const runtimeExecutable = workspace.pathToJest;
+
+  // "[^"\\]*                 Matches a quote followed by zero or more
+  //                          characters that are not quotes or backslashes
+  // (?:\\[\S\s][^"\\]*)*"    Matches a backslash followed by any character,
+  //                          followed by zero or more characters that are not
+  //                          quotes or backslashes ('[\S\s]' is used instead of
+  //                          '.' so that newlines are also matched). This
+  //                          ensures that escaped quotes are handled correctly
+  // (?:\\\s(?!\s+)|\S)       Matches a backslash followed by a single
+  //                          whitespace character, or a non-whitespace
+  //                          character
   const parameters = runtimeExecutable.match(
-    /(?:"[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|(?:\\\s|\S))+/g
+    /(?:"[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|(?:\\\s(?!\s)|\S))+/g
   ) || [''];
   const command = parameters[0];
   const initialArgs = parameters.slice(1);
--- a/src/Process.js
+++ b/src/Process.js
@@ -27,7 +27,9 @@ export const createProcess = (
   // as they can only be the first command, so take out the command, and add
   // any other bits into the args
   const runtimeExecutable = workspace.pathToJest;
-  const parameters = runtimeExecutable.split(' ');
+  const parameters = runtimeExecutable.match(
+    /(?:"[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|(?:\\\s|\S))+/g
+  ) || [''];
   const command = parameters[0];
   const initialArgs = parameters.slice(1);
   const runtimeArgs = [].concat(initialArgs, args);
--- a/src/Process.js
+++ b/src/Process.js
@@ -9,7 +9,6 @@
 
 import {ChildProcess, spawn} from 'child_process';
 import ProjectWorkspace from './project_workspace';
-import type {SpawnOptions} from './types';
 
 /**
  * Spawns and returns a Jest process with specific args
@@ -18,11 +17,7 @@ import type {SpawnOptions} from './types';
  * @returns {ChildProcess}
  */
 // eslint-disable-next-line import/prefer-default-export
-export const createProcess = (
-  workspace: ProjectWorkspace,
-  args: Array<string>,
-  options?: SpawnOptions = {}
-): ChildProcess => {
+export const createProcess = (workspace: ProjectWorkspace, args: Array<string>): ChildProcess => {
   let runtimeExecutable = workspace.pathToJest;
 
   if (args.length) {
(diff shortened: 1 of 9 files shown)


src/Process.js
```diff
@@ -27,7 +27,9 @@ export const createProcess = (
   // as they can only be the first command, so take out the command, and add
   // any other bits into the args
   const runtimeExecutable = workspace.pathToJest;
-  const parameters = runtimeExecutable.split(' ');
+  const parameters = runtimeExecutable.match(
+    /(?:"[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|(?:\\\s|\S))+/g
```
connectdotz: I think I get what you are trying to do, but it is a bit complex for the mere mortal: nested non-capture-groups, [\S\s] (isn't white-space + non-white-space = well, everything?)... I was having a hard time following, wondering if it is possible to simplify it, came up with this 
```
/(\S*".*?")|(\S*'.*?')|(\S*\\\s\S+)|\S+/g
```
there are really just 3 patterns we want to capture beyond the non-white-space: single, double quotes and backslash escape... tested with the jest and the following, both seem to work:

```
"/a/b is a directory/c" --without /x/y/z useThis="I am a 'nested single quote' test" single='this is a single quote'  backslash=a\\ space/b/c 
```

however, the nested quotes cause problems for both regex patterns:
```
--nestedQuote="before \\"after this\\" end" --nestedSingleQuote='before \\'after this\\' end'
```
while this can be easily fixed with lookbehind (I am sure you can also fix your regex pattern), it does start to get more complicated and ugly... I can be talked into not to support it as I don't see how this will occur in jest command line...
```
/(\S*".*?(?<!\\)")|(\S*'.*?(?<!\\)')|(\S*\\\s\S+)|\S+/g
```

anyway, just to throw some idea out there to see if we can get it simplified a bit...

BTW, to make future maintenance easier, how about adding some comments to explain the regex patterns?

connectdotz (commented):
This is one of the high priority bugs on our todo list for a while, thanks for tackling it 👍 

For the most part, the regex pattern worked, we could merge as it is but figure might as well take the opportunity to see if we can simplify it a bit to make future maintenance easier...

While the code change is trivial, the impact is not, just see how many issues it could resolve! I would encourage you do update the CHANGELOG.md to record this change, there will be many users want to thank you for it! 😄

connectdotz:
just to add one more note so we won't forget. once this PR is merged, we should patch the vscode-jest to make sure the default jest.pathToJest path is quoted so space will not be an issue and people don't need to override jest.pathToJest to work around this issue anymore...

src/Process.js
omjadas: The reason for `[\S\s]` is that it will match newlines, whereas `.` will not. This is in case of something like this

```sh
"/a/b is a directory/c" --without /x/y/z useThis="I am a 'nested 
single quote' test" single='this is a single quote'  backslash=a\\ space/b/c
```

Also, I'm pretty sure that 

```sh
--nestedQuote="before \\"after this\\" end"
```

should be two matches. Whereas

```sh
--nestedQuote="before \"after this\" end"
```

should be one match.

I noticed that the regex you suggested split

```sh
"/a/b is\" adirectory/c"
```

into two matches. It's unlikely that someone would have a quote in their path, but it is possible.

Not sure how to simplify it and still support these cases.

connectdotz:
> I noticed that the regex you suggested split
> "/a/b is\" adirectory/c"
> into two matches. 

hmmm, did you use the first regex pattern that I stated doesn't support nested quote? or the 2nd pattern that does: `/(\S*".*?(?<!\\)")|(\S*'.*?(?<!\\)')|(\S*\\\s\S+)|\S+/g`  ? because the later seems to work just fine, ins't it?

> The reason for [\S\s] is that it will match newlines

I see. I think it is pretty safe not to support the newline for our use case...

---

In any case, optimization or not, the most important thing is that the process needs to run... I tested this PR with `vscode-jest` on mac and looks like we have another issue...The arguments are indeed no longer split on space, but the spawned process failed: 

for example,  for the settings: `"jest.pathToJest": "yarn test --config='/Users/x/dir with space/jest.config.js'",`. 

Before spawning, the arguments looked all right to me: 

`spawning process with command=yarn, args=test,--config='/Users/x/dir with space/jest.config.js',--testLocationInResults,--json,--useStderr,--outputFile,/var/folders/vg/yw9mltwd1mj_29xbvhwywc980000gn/T/jest_runner_jest_editor_support.json,--watch,--no-color`

but the spanwed process error out:
```
$ jest '--config='\''/Users/vsun/github/sandbox/dir with space/jest-editor-support/jest.config.js'\' --testLocationInResults --json --useStderr --outputFile /var/folders/vg/yw9mltwd1mj_29xbvhwywc980000gn/T/jest_runner_jest_editor_support.json --no-color

Usage: jest [--config=<pathToConfigFile>] [TestPathPattern]
...
```
Apparently `child_process.spawn` is quoting the whole config argument as a string, and escaping the inner quotes that jest simply doesn't understand...

also tried (single/double) escape space hoping to by-pass quote problem above but without success...

thoughts?

omjadas:
I think setting `shell: true` might fix it. I can give it a go over the weekend, don't have access to macOS though.

omjadas:
Setting `shell: true` seems to fix it for the case that you have given. Is there a reason that is only being set to true for Windows in `vscode-jest` at the moment?

seanpoulter:
To say it again, thank you @omjadas for tackling this bug! 👍 

--

> Is there a reason that is only being set to true for Windows in `vscode-jest` at the moment?

That change sounds familiar to me. I think the change was introduced to solve a problem that was only on Windows. We didn't think about non-Windows systems or have as much capacity to test it. Here's [the PR](https://github.com/jest-community/vscode-jest/pull/248/files#diff-2b48e411c0f9ef17b99cba1f59a5012f) that added it.

--


If we're still having problems trying to be clever and parse a user-provided string into the `command` and `args` for [`spawn(command: string, args: string[], ...)`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options), I'd like to make a few suggestions:
* We should extract this transformation into a small testable function.
* We should capture all your great examples in the PR comments in our tests.

connectdotz:
> Is there a reason that is only being set to true for Windows in vscode-jest at the moment?

the shell option comes with its overhead (the shell) and [security risk](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options):
> If the shell option is enabled, do not pass unsanitized user input to this function. Any input containing shell metacharacters may be used to trigger arbitrary command execution.

Therefore, no reason to use it unless we absolutely need it...

Having say that, if we can't find a better alternative and must use shell, I think it might be all right, provided we do some tests:
1. the security risk is probably acceptable since the processes are executed in users' own environment.
1. we can probably remove the command args parsing all together and just pass the full command string to spawn (test needed)
1. not sure how much overhead with the extra shell, especially in multi-root workspace where we alreadh spawn quite a few jest processes... (test needed)

connectdotz:
@omjadas not sure if you have a chance to test the shell option yet, I did a quick and dirty test with your suggestion: enable "shell" option, then removing all parsing/splitting logic, appending the extra args at the end of the `runtimeExecutable`, and simply invoke spawn like `spawn(commandWithArgs, [], spawnOptions)`, it seems to work just fine. 

do you have any concerns? do you think you will have some time to try this new approach?

omjadas:
No immediate concerns jump out to me. I'll be able to do some testing over the weekend.

omjadas:
I've done some testing and I was not able to determine a difference in performance when setting the shell option in multi-root workspaces.

Did we want to remove the options argument, or leave it in case we want to pass other options to spawn in the future?

connectdotz:
> Did we want to remove the options argument, or leave it in case we want to pass other options to spawn in the future?

I think we can leave it and maybe even check if the options.shell is set explicitly to "false", we could override the default "yes" as a way to turn-off the shell... but I admit that I can't think of an actual use case, other than testing and emergency workaround, if needed...

---

@omjadas  btw, did you get a chance to test this approach on windows? 

@seanpoulter and @stephtr, this is sort of a high impact change, do you guys have anything else to add? To be safe, it would be great if we can all test it in our local build for a few days before merge...

omjadas:
@connectdotz I was able to test with both windows and linux.

If we want to be able to explicitly set the shell option to false, should we re-add the arg splitting on spaces in that case?

connectdotz:
> If we want to be able to explicitly set the shell option to false, should we re-add the arg splitting on spaces in that case?

great point, this changed my mind, let's just remove the options. Please also clean up types such as `index.d.ts`.

omjadas:
I've removed spawnOptions and also the shell option from a few other places it occurred

src/Process.js
```diff
@@ -18,24 +17,17 @@ import type {SpawnOptions} from './types';
...
+
+  if (args.length) {
+    runtimeExecutable += ` ${args.join(' ')}`;
+  }
 
   // If a path to configuration file was defined, push it to runtimeArgs
   if (workspace.pathToConfig) {
-    runtimeArgs.push('--config');
-    runtimeArgs.push(workspace.pathToConfig);
+    runtimeExecutable += ' --config ';
```
connectdotz: while using string `+=` worked, one has to be mindful about adding space separators manually. It might be simpler to just push all command and arguments into an array, then use the `array.join(' ')` to construct the final command line?

connectdotz (commented):
looks good to me with just a minor suggestion. I have tested this in my repo for a few days, it seems to work fine, although we do need some other PR to fix the default `pathToJest` function, but that is beyond the scope of this PR.

src/Process.js
omjadas: thanks for the suggestion

PR #24, opened 2019-11-22
Add --reporters support

jmarceli:
`--reporters` option support added in order to avoid workarounds in https://github.com/jest-community/vscode-jest/pull/513/

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -79,7 +79,9 @@ export default class Runner extends EventEmitter {
       args.push('--no-color');
     }
     if (this.options.reporters) {
-      args.push('--reporters', this.options.reporters);
+      this.options.reporters.forEach(reporter => {
+        args.push('--reporters', reporter);
+      });
     }
 
     this.debugprocess = this._createProcess(this.workspace, args);
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -78,6 +78,9 @@ export default class Runner extends EventEmitter {
     if (this.options.noColor === true) {
       args.push('--no-color');
     }
+    if (this.options.reporters) {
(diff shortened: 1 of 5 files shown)


nicu (approved):

src/types.js
```diff
@@ -20,6 +20,7 @@ export type Options = {
   noColor?: boolean,
   testNamePattern?: string,
   testFileNamePattern?: string,
+  reporters?: string,
```
connectdotz: I think you might want to take an array, see [jest cli](https://jestjs.io/docs/en/cli#--reporters).

also please update `index.d.ts`.

src/Runner.js
```diff
@@ -78,6 +78,9 @@ export default class Runner extends EventEmitter {
     if (this.options.noColor === true) {
       args.push('--no-color');
     }
+    if (this.options.reporters) {
+      args.push('--reporters', this.options.reporters);
```
connectdotz: see [jest cli](https://jestjs.io/docs/en/cli#--reporters), `this.options.reporters` probably should be an array and we will need to pass each with its own `--reporters` flag.

connectdotz (changes requested):
thanks, @jmarceli for adding this great feature 👍 

Just a few comments below, which should be pretty straightforward to address. Hopefully, we can cut a minor release with this PR soon so it can be used by your other PR in vscode-jest.

src/types.js
jmarceli: Done. Thanks for pointing that out.

src/Runner.js
jmarceli: Good idea. Done.

PR #26, opened 2019-11-28
Add --bail and --runInBand

jmarceli:
Using `--bail` and `--runInBand` may sometimes improve DX and possibly mitigate performance issues. It might be the first approach to address issues like https://github.com/jest-community/vscode-jest/issues/455

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -83,6 +83,12 @@ export default class Runner extends EventEmitter {
         args.push('--reporters', reporter);
       });
     }
+    if (this.options.runInBand === true) {
+      args.push('--runInBand');
+    }
+    if (this.options.bail === true) {
+      args.push('--bail');
+    }
 
     this.debugprocess = this._createProcess(this.workspace, args);
     this.debugprocess.stdout.on('data', (data: Buffer) => {
--- a/src/__tests__/runner.test.js
+++ b/src/__tests__/runner.test.js
@@ -288,6 +288,32 @@ describe('Runner', () => {
       const lastIndex = args.lastIndexOf('--reporters');
       expect(args[lastIndex + 1]).toBe(expected[1]);
     });
+
+    it('calls createProcess with --runInBand arg when provided', () => {
+      const expected = '--runInBand';
+
+      const workspace: any = {};
+      const options = {runInBand: true};
+      const sut = new Runner(workspace, options);
+      sut.start(false);
+
+      const args = (createProcess: any).mock.calls[0][1];
+      const index = args.indexOf(expected);
+      expect(index).not.toBe(-1);
+    });
+
+    it('calls createProcess with --bail arg when provided', () => {
+      const expected = '--bail';
+
+      const workspace: any = {};
+      const options = {bail: true};
+      const sut = new Runner(workspace, options);
+      sut.start(false);
+
+      const args = (createProcess: any).mock.calls[0][1];
+      const index = args.indexOf(expected);
+      expect(index).not.toBe(-1);
+    });
   });
 
   describe('closeProcess', () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
+* Add `--runInBand` and `--bail` options support - [@jmarceli](https://github.com/jmarceli)
 
 -->
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,6 +18,8 @@ export interface Options {
   testNamePattern?: string;
   testFileNamePattern?: string;
   reporters?: string[];
+  runInBand?: boolean;
+  bail?: boolean;
 }
 
 export class Runner extends EventEmitter {
--- a/src/types.js
+++ b/src/types.js
@@ -21,6 +21,8 @@ export type Options = {
   testNamePattern?: string,
   testFileNamePattern?: string,
   reporters?: string[],
+  runInBand?: boolean,
+  bail?: boolean,
 };
 
 /**


src/Runner.js
```diff
@@ -83,6 +83,12 @@ export default class Runner extends EventEmitter {
         args.push('--reporters', reporter);
       });
     }
+    if (this.options.runInBand === true) {
+      args.push('--runInBand');
+    }
+    if (this.options.bail === true) {
+      args.push('--bail');
+    }
```
seanpoulter, line 91: Wow, we are strict with our truth values. 🤣

src/__tests__/runner.test.js
```diff
@@ -288,6 +288,32 @@ describe('Runner', () => {
...
+
+      const workspace: any = {};
+      const options = {bail: true};
+      const sut = new Runner(workspace, options);
+      sut.start(false);
+
+      const args = (createProcess: any).mock.calls[0][1];
+      const index = args.indexOf(expected);
+      expect(index).not.toBe(-1);
+    });
```
seanpoulter, line 316: I appreciate that you've maintained the coding style for the tests before this.

This is a non-blocking comment about the style/readability. I've been away from the repo and Jest long enough I had to look up the mock.calls type. I think some array destructuring and using the `toContain` matcher would make these two a bit more readable at a glance (for someone like me who's been absent):
```js
      const [_workspace, args] = (createProcess: any).mock.calls[0];     
      expect(args).toContain('--runInBand'); // or .toContain(expected);
```

seanpoulter (approved):
Thank you @jmarceli and sorry for the painfully slow turnaround. I'm OK to merge this as is. We'll need @connectdotz (or @orta) for a deployment though.

stephtr:
First of all thanks for this PR!
As @connectdotz [already mentioned in #455](https://github.com/jest-community/vscode-jest/issues/455#issuecomment-564428051), we should think about just adding flags to jest-editor-support, which will be used by vscode-jest (or similar tools).
For `bail` I currently can't see the use case, since in VS Code at least I would like to have an overview over which test is passing and which one is failing. If we would abort Jest at the first failing test, we wouldn't have any information available about the remaining ones. Besides that, it only shortens runtime/CPU consumption when having failing tests.
In contrast to that `runInBand` would make sense for me to include, since it apparently can speed up jest in resource limited environments. Since it depends on the environment and not on the individual project, I would be also in favor of adding a setting to vscode-jest controlling (provided that the performance boost is noticeable for a larger audience) this flag.

connectdotz:
@seanpoulter nice to see you back!

Regarding this issue: my gut feeling is no need to add these flags because:
1. users are already able to customize with these flags 
  users can easily achieve the same effect, and maybe more, by just add these flags in their `jest.pathToJest` or package.json script. 

2. it doesn't save any user effort 
  Unless we want all users to run with these flags, otherwise, If the user had to customize their env to enable this behavior anyway, wouldn't it be just as easy to customize their cmd line as described above?  

3. it may or may not address the performance issue
  Why stop at `--runInBand`? How about `--maxWorkers` and maybe more? The good news is we don't have to stand in the way! Users can do all these today with the customization mentioned above.

seanpoulter:
> @seanpoulter nice to see you back!

Nice to see you two too! My apologies for not having more time to contribute. I don't get to work with Jest or the VS Code extension much at all.

--

I'll admit my initial reaction was pretty similar to you @connectdotz. It seems strange that we're so selective with what we've included in our wrapper/facade here. I can't think of a use case where I'd want to support another copy of the API or the Jest documentation here or in any caller. 😓 

Do we close it until we have the performance warning resolved?

stephtr:
> users are already able to customize with these flags

In my opinion the main issue with `jest.pathToJest` is that it isn't possible to separate between flags which are set on a project basis and flags that are set on a machine basis.
For example let's assume that `pathToJest` is set on a project basis, there is no way to append `--maxWorkers` on old, slow machines. One could append it to the project's `pathToJest` settings, but that would also affect fast machines when working on the project.

jmarceli:
Hi, I can agree that those new flags are not required if we can address them using `jest.pathToJest` (I didn't think about that).

Regarding performance, it would be nice if anyone else can try running their test suites with and without `--runInBand` flag. In every project and on every machine I can run my tests (including CI servers) it was always faster to run tests with `--runInBand` flag. It would be nice if you can then look at https://github.com/jest-community/vscode-jest/issues/455 which is an issue connected with the performance of vscode-jest.

seanpoulter:
> In every project and on every machine I can run my tests (including CI servers) it was always faster to run tests with --runInBand flag

Wow, that's huge performance feedback @jmarceli!

<br/>

> It would be nice if you can then look at jest-community/vscode-jest#455 which is an issue connected with the performance of vscode-jest.

It sounds like there'd be value in some A/B testing and something that the Jest team may have already investigated. @thymikee confirms your observation that `--runInBand` is suggested for CI deployments in facebook/jest#3765, especially when tests are running with more than one version of Node. I'm curious if anyone's already found a way to optimize the number of workers.

After some brief searching:

* @thymikee was interested: https://github.com/facebook/jest/issues/5989#issuecomment-381434078
* @SimenB investigated: https://github.com/facebook/jest/issues/5989#issuecomment-417368228
* Here's a blog with someone's arbitrary recommendation for maxWorkers: https://www.peterbe.com/plog/ideal-number-of-workers-in-jest-maxWorkers

There's a lot of valuable points from their conversation:
* Limiting the number of workers to the number of available CPUs adds other challenges
* `runInBand` may not be the most performant the first time we run the whole test suite, but may be ideal running in watch mode.
* Solving the "optimal" configuration depends on the system resources and availability
* It seems like the Jest team's solution is to help set "appropriate" defaults and educate users to do the rest.

I think we've got a hypothesis or two that we could test. It'd be lovely if we could run some A/B tests and collect performance data from our users to make an informed decision about it. 🤔

Issue #29, opened 2020-01-21
parse does not find name for test.each block


firsttris:
```
      test.each([
        [1, 1, 2],
        [1, 2, 3],
        [2, 1, 3]
      ])('wtf', (a, b, expected) => {
        expect(a + b).toBe(expected);
      });
```
https://jestjs.io/docs/en/api#testeachtablename-fn-timeout
(note: its not possible to put a test.each inside a "it" or a "test" )

parser outputs a warning if he detects such a block.
```
  console.warn node_modules/jest-editor-support/build/parsers/typescript_parser.js:96
    **NamedBlock but no name found for it tsNode= NodeObject** {
      pos: 442,
      end: 526,
      flags: 0,
      modifierFlagsCache: 0,
      transformFlags: 0,
      parent: undefined,
      kind: 195,
      expression: NodeObject {
        pos: 442,
        end: 459,
        flags: 0,
        modifierFlagsCache: 0,
        transformFlags: 0,
        parent: undefined,
        kind: 193,
        expression: IdentifierObject {
          pos: 442,
          end: 454,
          flags: 0,
          modifierFlagsCache: 0,
          transformFlags: 0,
          parent: undefined,
          kind: 75,
          escapedText: 'test'
        },
        questionDotToken: undefined,
        name: IdentifierObject {
          pos: 455,
          end: 459,
          flags: 0,
          modifierFlagsCache: 0,
          transformFlags: 0,
          parent: undefined,
          kind: 75,
          escapedText: 'each'
        }
      },
      questionDotToken: undefined,
      arguments: [
        NodeObject {
          pos: 460,
          end: 525,
          flags: 0,
          modifierFlagsCache: 0,
          transformFlags: 0,
          parent: undefined,
          kind: 191,
          multiLine: true,
          elements: [Array]
        },
        pos: 460,
        end: 525
      ]
    }
```

firsttris:
the parsed object for this block:
```
{
                                        "type": "it",
                                        "file": "filename...",
                                        "start": {
                                             "column": 7,
                                             "line": 22
                                        },
                                        "end": {
                                             "column": 37,
                                             "line": 26
                                        }
                                   },
                                   {
                                        "type": "expect",
                                        "file": "filename...",
                                        "start": {
                                             "column": 9,
                                             "line": 27
                                        },
                                        "end": {
                                             "column": 37,
                                             "line": 27
                                        },
                                        "children": [
                                             {
                                                  "type": "expect",
[...]

PR #30, opened 2020-01-24
update babylon to babel/parser 7.x

firsttris:
Hey jest-community,

i update babylon to babel/parser 7.x as suggested by @seanpoulter https://github.com/jest-community/jest-editor-support/issues/29

now the typescriptt_parser.js is actually no longer relevant and could possibly be removed. also the typescript dependency in package.json.
since the babel_parser.js is now able to parse typescript as well.

~~i did not delete typescript_parser.js and corresponding test yet. (even if it would no longer be used after this PR)~~

i add a addtional test - babel_typescript_parser.test.js which executes babel_parser in typescript mode.

i checked that all tests are green, and fixed a few eslint issues.

if you are happy with this PR, i would proceed update the jest parser to new features like test.each.

regards
Tristan


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Bug-fixes within the same version aren't needed
 
 -->
 
+### 27.1.1
+* replace babylon and typescript parser with @babel/parser 7.x
+
```
seanpoulter: It has been a while, but I think we need to move this up under the `## Master` heading. Changes that haven't been released yet are kept hidden until they're released. I think the build tooling will move and rename the heading with the release version for us (unless it's always been @orta or @connectdotz 😇). I have no idea what would happen if there's a new version heading. 🤔 

```suggestion
## Master
* Replace babylon and typescript parsers with @babel/parser 7.x

```

fixtures/parser_tests.js
```diff
@@ -3,26 +3,21 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
```
seanpoulter, line 6: Why's the style changing? There's a few changes in this file I wouldn't expect, but I'll assume it's a code formatting thing... 🤨

fixtures/parser_tests.js
```diff
@@ -372,7 +330,7 @@ function parserTests(parse: (file: string) => ParserResult) {
         22,
         7,
         23,
-        18,
+        18
```
stephtr, line 333: Is there any reason for removing trailing commas?

src/parsers/babel_parser.js
```diff
@@ -10,28 +10,27 @@
...
-
-import {File as BabylonFile, Node as BabylonNode, parse as babylonParse} from 'babylon';
+import {File as BabelFile, Node as BabelNode} from '@babel/types';
+import * as parser from '@babel/parser';
 import type {ParsedNodeType} from './parser_nodes';
 import {NamedBlock, ParsedRange, ParseResult, ParsedNode} from './parser_nodes';
 
-// eslint-disable no-underscore-dangle
-const _getASTfor = (file: string, data: ?string): [BabylonFile, string] => {
+const _getASTfor = (file: string, data: ?string, options: parser.ParserOptions): [BabelFile, string] => {
```
stephtr: Since we are using flow, shouldn't the it be `?parser.ParserOptions`? To me `options` seems to be optional...

src/parsers/babel_parser.js
```diff
@@ -10,28 +10,27 @@
...
 };
 
-export const getASTfor = (file: string, data: ?string): BabylonFile => {
+export const getASTfor = (file: string, data: ?string): BabelFile => {
   const [bFile] = _getASTfor(file, data);
   return bFile;
 };
 
-export const parse = (file: string, data: ?string): ParseResult => {
+export const parse = (file: string, data: ?string, options: parser.ParserOptions): ParseResult => {
```
stephtr: ...we could make `options` optional here, too. That way the signature of `parse` would be backward compatible.

stephtr (commented):
Why would we want to keep `typescript_parser.js` but not `babylon_parser.js`?

src/parsers/babel_parser.js
firsttris: absolutely, after an optional param cannot be a required param.

fixtures/parser_tests.js
firsttris: i double checked and it seems to be defined in the eslint / prettier config like this. i just jumped into the file and autosave triggered autoFixOnSave which triggered eslint & prettier to fix the linting / formatting according to the defined config.

If i add this trailing comma, vscode tells me that there is a eslint issue.

src/parsers/babel_parser.js
firsttris: absolutely, after an optional param cannot be a required param.

connectdotz:
@firsttris thanks for the PR, before diving into the code, here is a [conversation](https://github.com/facebook/jest/pull/6949#pullrequestreview-155755901) over a year ago that you might find relevant. 

IMHO, the most uncertain part of this migration is the plugin list. We used to load all the plugin (yeah, that is a bit crazy), now with an explicit plugin list, I am not sure if we will encounter issues that the user plugin list != our parser's. Will that be an issue for parsing their code? If it does, do we need to consider allow users to pass in their plugin list or even reading their babel config? (a can of worms...) Thoughts?

firsttris:
reading the users babel config sounds complex to me . big fan of kiss principle. 

if the parser supports more features that then user is using everything works well.
if the parser does not support features the user is using it will probably not parse the code and generate errors.

if adding all plugins has worked before.. why not continue this way?

i tried to add as many plugins as possible.
there was no measurable speed difference when adding all those plugin.
(did a tests but exacly the same ms when executing tests)

some plugins:
- decorators
- pipelineOperator
did not work

they outputed this errors:
The 'decorators' plugin requires a 'decoratorsBeforeExport' option, whose value must be a boolean. If you are migrating from Babylon/Babel 6 or want to use the old decorators proposal, you should use the 'decorators-legacy' plugin instead of 'decorators'.
'pipelineOperator' requires 'proposal' option whose value should be one of: 'minimal', 'smart', 'fsharp'

i actually added them as an array of strings?

also removed the typescript_parser as @stephtr suggested.

connectdotz:
> if adding all plugins has worked before.. why not continue this way?

reasonable starting point.

> some plugins ... did not work

see [discussion](https://github.com/facebook/jest/pull/7016#pullrequestreview-157669721)

> i actually added them as an array of strings?

should be... but we should probably create some fixtures with some of these features, such as optional-chaining, dynamic import etc, to make sure they can be parsed.

stephtr:
> Will that be an issue for parsing their code? If it does, do we need to consider allow users to pass in their plugin list or even reading their Babel config?

There could be situations where Babel won't work with our configuration, but I'm quite sure that also the Babylon parser (as it is currently used) would have failed for them, it therefore wouldn't be a regression.

fixtures/parser_tests.js
```diff
@@ -3,26 +3,21 @@
...
+function parserTests(parse: (file: string) => ParseResult) {
   const assertBlock = (block, start, end, name: ?string = null) => {
     expect(block.start).toEqual(start);
     expect(block.end).toEqual(end);
     if (name) {
       expect(block.name).toEqual(name);
     }
   };
+  const assertBlock2 = (block, sl: number, sc: number, el: number, ec: number, name: ?string = null) =>
+    assertBlock(block, {column: sc, line: sl}, {column: ec, line: el}, name);
```
seanpoulter, line 20: Thanks for cleaning up as you go! 👍

fixtures/parser_tests.js
```diff
@@ -315,15 +275,13 @@ function parserTests(parse: (file: string) => ParserResult) {
...
+    multiple lines`
       );
       assertBlock(e1, {column: 5, line: 24}, {column: 32, line: 24});
     });
 
     test(`edge case: should not fail`, () => {
-      const dBlock = parseResult.describeBlocks.filter(
-        b => b.name === 'edge case: should not fail',
-      )[0];
+      const dBlock = parseResult.describeBlocks.filter(b => b.name === 'edge case: should not fail')[0];
```
seanpoulter, line 284: This is unrelated to your change @firsttris. Would you welcome a PR to give these 5+ `[0]`s some sort of name for context, @connectdotz and @stephtr? I can't remember what they're supposed to be. 😆 

In this test in particular, I __think__ we're asserting that:
* the first test suite matching our predicate has 2 test suites
* the first test has one expect statement
* we've correctly parsed the location in the file(s)

I'm struggling and I may have seen this before. ☺️

src/parsers/babel_parser.js
```diff
@@ -171,3 +170,34 @@ export const parse = (file: string, data: ?string): ParseResult => {
...
+  'partialApplication',
+  'throwExpressions',
+  'topLevelAwait',
+];
+
+// Its not possible to use the parser with flow and typescript active at the same time
+export const parseJs = (file: string, data: ?string): ParseResult =>
+  parse(file, data, {plugins: [...additionalPlugins, 'flow']});
+export const parseTs = (file: string, data: ?string): ParseResult =>
+  parse(file, data, {plugins: [...additionalPlugins, 'typescript']});
```
seanpoulter: Thanks for the comment explaining the problem with flow and typescript at the same time.

It looks like we used to get all the plugins using '*'. How'd you decide which plugins to use here? Any tips for us maintaining it later? 💃

src/__tests__/runner.test.js
```diff
@@ -403,6 +403,13 @@ describe('events', () => {
     expect(close).toBeCalled();
   });
 
+  it('emits debuggerProcessExit when process close', () => {
+    const close = jest.fn();
+    runner.on('debuggerProcessExit', close);
+    fakeProcess.emit('close');
+    expect(close).toBeCalled();
+  });
+
```
seanpoulter, line 438: Yay! 👍

seanpoulter (approved):
Hurrah! I think this is a big step in the right direction! Thank you! There's only one question that @connectdotz and/or @stephtr didn't seem to already cover - is the CHANGELOG.md OK like that? I'll approve it now so we're not stuck with me being sloooooooooow. 😪 

You've already had the discussion about use plugins. We'll probably want a seam / way for the user to provide options, but we can add options to the `parse` functions in another PR.

seanpoulter:
Oh!! I am tired and forgot to ask about Yarn. Does Yarn update package-lock.json now too or did you use `npm`? I'm not bothered by the politics of it, I just want to make sure we're all on board with our build process (which seems to be undocumented).

fixtures/parser_tests.js
stephtr: I think in this case it just checks whether there is any `describeBlock` with the given name. One probably should use `describeBlocks.find(...)` instead.

fixtures/parser_tests.js
stephtr: I think it's due to eslint and/or prettier. If so, I'm just wondering, that they didn't kick in earlier.

fixtures/parser_tests.js
seanpoulter: Just the function we need. 👍

src/parsers/babel_parser.js
connectdotz: wildcard plugin list no longer supported in babel-7. The current decision is to include them *all*, or as much as possible. Yes as new plugins become available and our users start to use them, we might need to come back here to add them or consider a way for users to pass their list of plugins. 

We could deal with that later... if it happened as infrequent as it was then manually maintaining a full list might not be too bad. But I think **we should call out the plugins we currently support in the README file.**

src/parsers/babel_parser.js
```diff
@@ -171,3 +170,34 @@ export const parse = (file: string, data: ?string): ParseResult => {
...
+  'logicalAssignment',
+  'nullishCoalescingOperator',
+  'numericSeparator',
+  'objectRestSpread',
+  'optionalCatchBinding',
+  'optionalChaining',
+  'partialApplication',
+  'throwExpressions',
+  'topLevelAwait',
+];
```
connectdotz: I am not sure not adding decorators and pipelineOperator is better than adding with some defaults like legacy? Either way, we should document this list in the README file and give a bit of explanation of why and the limit etc.

src/parsers/index.js
```diff
@@ -8,8 +8,7 @@
  */
 
 import {ParseResult} from './parser_nodes';
-import {parse as parseJs} from './babylon_parser';
-import {parse as parseTs} from './typescript_parser';
+import {parseJs, parseTs} from './babel_parser';
```
connectdotz: this is not in your change, but while you are right here, can you add support for `mjs` file extension as seen  in jest-community/vscode-jest#551, particularly [here](https://github.com/jest-community/vscode-jest/issues/551#issuecomment-579529020). Hopefully, the optional-chaining issue will be automatically resolved once this PR is merged.

connectdotz (changes requested):
@stephtr and @seanpoulter already covered many grounds, I will just add a few relatively simple requests. I think we are quite close to merging this!

P.S. I assume you are going to do the actual parser change for the `test.each` in a separate PR, this is just to switch over to the babel-7 world, right?

src/__tests__/runner.test.js
firsttris: i needed to increase my coverage :) coveralls didn't let me pass

src/parsers/index.js
firsttris: i look into it

CHANGELOG.md
firsttris: thx ! :)

fixtures/parser_tests.js
firsttris: the boy scout rule

firsttris:
> are going to do the actual parser change for the `test.each` in a separate PR, this is just to switch over to the babel-7 world, right?

yes, in those enterprises they told to always keep my PR's small :)

i don't wanted to extend two parsers.

firsttris:
> Oh!! I am tired and forgot to ask about Yarn. Does Yarn update package-lock.json now too or did you use `npm`? I'm not bothered by the politics of it, I just want to make sure we're all on board with our build process (which seems to be undocumented).

i think i used npm, as far as i know package-lock.json is only created by npm

src/parsers/babel_parser.js
firsttris: i will add a list to the readme, but in general there is little information about this package in the readme. would be great if there would be a short summary about the methods this package is exposing.

CHANGELOG.md
firsttris: is it correct now?

src/parsers/babel_parser.js
connectdotz: agree, let's start building up the README file

connectdotz:
> i think i used npm, as far as i know package-lock.json is only created by npm

I wonder if we allow both yarn and npm lock files, will we encounter a situation that developers will end up with different versions of the dependencies based on their pm preference?  Or otherwise we will ask developers to update both lock files by running boh yarn and npm? 🤔 Maybe it is best to standardize on the PM, which right now is yarn...

firsttris:
> > if adding all plugins has worked before.. why not continue this way?
> 
> reasonable starting point.
> 
> > some plugins ... did not work
> 
> see [discussion](https://github.com/facebook/jest/pull/7016#pullrequestreview-157669721)
> 
> > i actually added them as an array of strings?
> 
> should be... but we should probably create some fixtures with some of these features, such as optional-chaining, dynamic import etc, to make sure they can be parsed.

@connectdotz 
actually we could let the api consumer pass the addtionalPluginList as an addtional param of the parse method? maybe this would be a better option then maintaining the list in this package?

src/parsers/index.js
firsttris: implemented

src/parsers/babel_parser.js
firsttris: i added some info to the Readme, please verify. actually we could let the api consumer pass the addtionalPluginList as an addtional param of the parse method? maybe this would be a better option then maintaining the list in this package?

src/parsers/index.js
```diff
@@ -8,25 +8,24 @@
...
 
 /**
  * parse the test file by selecting proper parser based on the file extension.
  *
  * exception will be throw should the underlying parse failed.
  */
-export default function parse(file: string, data?: string, strict: boolean = false): ParseResult {
-  if (file.match(/\.tsx?$/)) {
-    return parseTs(file, data);
+export default function parse(filePath: string, serializedData?: string, strictMode?: boolean = false): ParseResult {
```
firsttris: i adjusted those names to make their purpose more clear..

src/parsers/babel_parser.js
firsttris: we could add addtionalPlugins as param to parse? then it would be more dynamic for the API user. also we would not have to adjust the package if a new proposal was released.

src/parsers/index.js
stephtr: Just a side remark: Since `strictMode` has a default value, it doesn't have to be nullable.

src/parsers/babel_parser.js
firsttris: please check how i imagine this could work https://github.com/jest-community/jest-editor-support/pull/30/files#diff-80745ee8575f2870ead6cd5b032a42d2

this way we would not need to hardcode all the plugins into this package. instead leave it up to the API consumer what to activate. we could also still adjust which plugins we include in this package and define them in readme.md https://github.com/jest-community/jest-editor-support/pull/30/files#diff-2427e229bfc5afdbd2cbbc3ed96a650fR174

what do you think? will i get the job? :bowtie:

src/parsers/babel_parser.js
```diff
@@ -171,3 +170,11 @@ export const parse = (file: string, data: ?string): ParseResult => {
...
   return parseResult;
 };
+
+export const plugins = ['jsx', 'classProperties'];
+
+// Its not possible to use the parser with flow and typescript active at the same time
+export const parseJs = (file: string, data: ?string, additionalPlugins: string[] = []): ParseResult =>
+  parse(file, data, {plugins: [...plugins, ...additionalPlugins, 'flow']});
+export const parseTs = (file: string, data: ?string, additionalPlugins: string[] = []): ParseResult =>
+  parse(file, data, {plugins: [...plugins, ...additionalPlugins, 'typescript']});
```
connectdotz: Allow customization of the plugin is great 👍 , but I think I would make it more "optional" by providing a solid default set of plugins but allow  **override** (customization might mean add or remove plugins). I think this way it provides the least interruption while adding flexibility.

README.md
```diff
@@ -10,7 +10,23 @@ The engine that allows editors to build on top of Jest.
...
+  strictMode: boolean = false,   
+  additionalPlugins?: stirng[]   
+)
+```
+Parse is a static Jest parser which uses Babel 7 and supports js,jsx,ts,tsx,mjs files.   
+
+- filePath = Path to the file you want to parse.
+- serializedData = Serialized data, will be used instead of the filePath if available (optional).
+- strictMode = If this option is activated the parser throws an exception if the filetype is not detected, defaults to false.
+- additionalPlugins = List of ECMAScript proposal [plugins](https://babeljs.io/docs/en/babel-parser#ecmascript-proposals-https-githubcom-babel-proposals). Pass an array of strings e.g. ```['dynamicImport']```. Please have a look at the list of [plugins](https://github.com/jest-community/jest-editor-support/blob/master/src/parsers/babel_parser.js#L174) activated by default.
```
connectdotz: Would be even more helpful to provide an example for plugins customization.

README.md
```diff
@@ -10,7 +10,23 @@ The engine that allows editors to build on top of Jest.
...
+## API
+```
+parse(   
+  filePath: string,   
+  serializedData?: string,   
+  strictMode: boolean = false,   
+  additionalPlugins?: stirng[]   
+)
+```
+Parse is a static Jest parser which uses Babel 7 and supports js,jsx,ts,tsx,mjs files.
```
connectdotz: minor point: maybe move all the js file together like: js, jsx, mjs, ts, tsx

src/__tests__/parsers/babel_parser_plugin.test.js
```diff
@@ -0,0 +1,50 @@
...
+    parse('fileName.js', codeTobeParsed, false, additionalPlugins);
+    expect(spy).toBeCalledWith(codeTobeParsed, parserConfigFlow);
+  });
+
+  it('created plugins for typescript file correctly', () => {
+    const spy = jest.spyOn(babelParser, 'parse');
+    parse('fileName.ts', codeTobeParsed, false, additionalPlugins);
+    expect(spy).toBeCalledWith(codeTobeParsed, parserConfigTypescript);
+  });
+});
```
connectdotz: not sure if I missed it but I didn't see any additional fixture for testing plugin specific syntax?

connectdotz (changes requested):
@firsttris lots of good change here 👍, a few notes:

- plugin customization: nice feature, but suggest we still have a solid default, more detail inline
- readme: great start, some minor comment inline, might need to be updated if plugin customization changed.
- npm support: personally I prefer us to simply standardize on `yarn`. But if you would like to add npm support that would be fine as long as there is a mechanism to sync the 2 lock files automatically so they won't get out of sync.

package.json
```diff
@@ -52,10 +52,10 @@
   "dependencies": {
     "@babel/traverse": "^7.6.2",
     "@jest/types": "^24.8.0",
-    "babylon": "^6.14.1",
+    "@babel/parser": "^7.8.3",
+    "@babel/types": "^7.8.3",
     "core-js": "^3.2.1",
-    "jest-snapshot": "^24.7.0",
-    "typescript": "^3.4.3"
```
rossknudsen, line 67: Can we move the `typescript` package to dev dependencies?

src/__tests__/parsers/babel_parser_plugin.test.js
firsttris: no you didn't missed them, i have not added any yet.
i thought about this topic, normally you would only test your package, not the 3rd party depdenencies the project is using.
the plugin features should have already been tested in the babel project.

i mean we could not even fix a issue, if some tests were red, since its not within this project.

connectdotz: I see where you are coming from. I think I am less concerned about "if babel implemented the plugin correctly" but "if the plugin list we passed in has the effect we anticipated"...

I played with the new `babel-parser_plugin.test.js` to try to answer my own question:
- comment out all additional plugins, it passed; ok, maybe all of our code fixures/snippets are just standard js that requires no additional plugin...?
- passed in a garbage plugin, it still passed. ok, maybe babel load plugin like a best-effort...?
- I then created some optional chaining code-snippet with empty additional plugin list, thinking for sure it should fail without passing the correct plugin, right? But surprise surprise, it still passed.

At this point, I feel there is something I don't understand or misunderstand about this plugins thing... maybe it is reading the babel.config.js locally? does that mean it will be impacted by users' env as well once we deployed, which might be good, or bad... How do tools like `vscode-jest` or yours decide what plugin list to pass in? If they can't decide then should all users now need to pass in their own list, even if they don't use babel? With the current implementation (minimal default), how disruptive the next release will be? Do we really need to get tangled with the rather complex user babel config/env?

Anyway, tests mean to surface issues, and I think without the simple additional tests I would not have discovered the unexpected behaviors and questions... in addition to tests, we should answer the questions above to make sure we fully understand what the community and users might face with these changes... thoughts?

package.json
firsttris: I will rebase and move typescript to Dev.

src/__tests__/parsers/babel_parser_plugin.test.js
firsttris: i don't have an answer to your questions yet but i appreciate that you are trying to understand it properly.  so please don't get my comment wrong.

just one side thought, which might be interesting (in terms of testing)

majestic doesn't seem to take this topic that seriously:
https://github.com/Raathigesh/majestic/blob/master/server/services/ast/parser.ts#L10

they just use 3 plugins => "jsx", "classProperties", "optionalChaining"

rossknudsen: Hey @connectdotz.  I had a look at this too and think that I've made some headway.  Have a look at this [babel documentation](https://babeljs.io/docs/en/next/babel-parser.html#plugins) that I found helpful (I thought I knew what the plugins did by the names - I wasn't always right).  That page will give you some useful snippets to test with later on.

Experimenting with `babel-parser_plugin.test.js`:
Hopefully I'm not pointing out the obvious, but this file is aimed at testing whether the correct plugin config is passed to the parser rather than whether that configured parser can parse the code correctly.  You can see this by the fact that the "code" it is asked to parse is on line 26:
```js
const codeTobeParsed = 'const bla = 1';
```
which as you can see is pretty simple JS and is not testing any of the experimental language features that the plugins are supposed to be supporting.  Which is probably why when you removed all the plugins, it still parsed correctly and the tests passed.

When I extended that line with snippets from the Babel documentation like this:
```js
const codeTobeParsed = `
//classPrivateProperties
class A { #b = 1; }

//classPrivateMethods
class B { #c() {} }

//doExpressions
var a = do { if (true) { 'hi'; } };
`;
```
I got test failures when the appropriate plugins were not enabled.  Hope that helps.

firsttris: exacly @rossknudsen this test was only to verfiy if the addtionalPlugins have been merged with the default plugins correctly.

since i have added this feature i thought it make sense to test it.

firsttris:
@rossknudsen just noticed with  ["allowJs": false](https://github.com/jest-community/jest-editor-support/blob/master/tsconfig.json#L7) in tsconfig its not possible to jump into the javascript files from typescript files with strg + click in vscode.
But probably there is a reason for "allowJs": false ...

src/parsers/babel_parser.js
firsttris: i provided a more solid list of default plugins. (for feature i see used more regularly).
With spreading those plugins together the consumer is able to add addtional plugins.
But i don't see why anyone would want to remove plugins, this would require a different implementation. maybe like: use the alternativePlugins instead of plugins if provided.

src/__tests__/parsers/babel_parser_plugin.test.js
firsttris: also did some research on this topic. 

```babel-parser_plugin.test.js``` is a good place to test the babel plugins.

i found that for example ObjectRestSpread works even if no plugin is activated. While classPrivateProperties or doExpressions do only work with plugins activated.

that is probably because some proposal are already on a later stage and work even without explicitly activating them.

ObjectRestSpread and OptionalChaining are already in stage 4.

does this answer your concerns @connectdotz ?

**Comment by rossknudsen** [comment]
@firsttris yeah it's because the JS files have Flow typings in them. TS compiler doesn't understand them. Which is why we're using Babel to transpile, as it understands everything. 

So the tsconfig is really only being used to typecheck the TS files and we want it to ignore the JS files.

**Comment by firsttris** [comment]
hey @connectdotz & jest-community, i did my best to implement your suggestions and change requests.

If you have additional change request please let me know. once i find time between my worktime, i will do my best to implement them.

really want to continue extend that parser with test.each support, i see many projects heavely using this and our addons are not supporting this feature.

From the user's perspective, that's a bug.
https://github.com/jest-community/vscode-jest/issues/427
https://github.com/firsttris/vscode-jest-runner/issues/55

**Comment by connectdotz** [comment]
hey @firsttris sorry for the long delay, was really busy this month, I will try to take a look this weekend, thanks for being patient.

File: src/__tests__/parsers/babel_parser_plugin.test.js (hunk 16 above)
Comment (connectdotz) [reply src/__tests__/parsers/babel_parser_plugin.test.js]: @firsttris yes it does! thanks for doing the research 👍 

I also dig into the babel-parser code today and found that it checked for only a small set (~11) of plugins, such as `logicalAssignment`, `classProperties`, and the document did mention

> The latest ECMAScript version enabled by default (ES2017).

This is great and does eliminate my biggest concern. Obviously, as the language evolves the extra plugin list will too, but as long as we keep the `babel-parser` up to date we should be able to support the majority of the use cases.

Comment (connectdotz) [reply src/__tests__/parsers/babel_parser_plugin.test.js]: Another take away from `babel-parser` code:  The logic to recognize and process these non-standard language features are already in the parser, the plugin strings are just like flags to allow processing the corresponding syntax if encountered in the content-to-parse.

For `jest-editor-support`, I don't see much downside to allow the parser to parse as much as it can. If the users choose to constraint it for their specific use case, which should be rare, they can certainly provide its plugin list to narrow and override the default.

there are 2 plugins require additional flags, if there is really no obvious winner (more commonly used), then whichever flag is fine, provided we document it so users know how to override it if need to.

There is no point to pass in "new" plugin because the babel-parser we packaged will not understand them anyway...

File: CHANGELOG.md [hunk 18]
```diff
@@ -4,7 +4,7 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
-
+* Replace babylon and typescript parsers with @babel/parser 7.x
```
connectdotz:
```suggestion
* Replace babylon and typescript parsers with @babel/parser 7.x - @firsttris 
```

src/parsers/babel_parser.js
```diff
@@ -171,3 +170,20 @@ export const parse = (file: string, data: ?string): ParseResult => {
...
+  'objectRestSpread',
+  'optionalChaining',
+  'topLevelAwait',
+];
+
+// Its not possible to use the parser with flow and typescript active at the same time
+export const parseJs = (file: string, data: ?string, additionalPlugins: string[] = []): ParseResult =>
+  parse(file, data, {plugins: [...plugins, ...additionalPlugins, 'flow']});
+export const parseTs = (file: string, data: ?string, additionalPlugins: string[] = []): ParseResult =>
+  parse(file, data, {plugins: [...plugins, ...additionalPlugins, 'typescript']});
```
connectdotz: giving the `plugins` here are just "flags" for the babel-parser, I suggest we add all of them, you can reference their list [here](https://github.com/babel/babel/blob/928b9f8c9518284eac6d0598633f2ec373fc6d0c/packages/babel-parser/typings/babel-parser.d.ts#L97).

Once we enabled everything, there is no point to add "additional"... the customization should be very rare, if any, and it could be simply an override.

connectdotz (changes requested):
hey @firsttris not sure if you are still interested in this PR or maybe we both are waiting for the other to take the next step...?

I see the major blocker is in how we invoke babel-parser, i.e. the options, as well as how we customize it... see the inline comments. 

This change will most likely impact the README, tests, and maybe others... let me know when you are happy with the changes, I will do a final review then

Sorry for such a long delay, hope you haven't given up...

firsttris:
hey @connectdotz to be honest is was not totally clear for me what tasks where missing to complete this PR.
the PR is really big, and its seems not really transparent in general on github.

to summarize

the open tasks are:

- we remove the option to pass additional plugins, instead we include all plugins
- update readme accordingly
- update changelog (add author)

do you see any additional changes to be done @connectdotz ?
(sorry if i missed something the PR is so big)

connectdotz:
@firsttris  your summary is right, I will take another look when you are done...

src/parsers/babel_parser.js
firsttris: it was not possible to add every plugin, but i added almost every plugin. i removed the ones which caused conflicts.

firsttris: i wonder if adding all those plugins is even necessary

firsttris:
did it

but travis now says

The command "eval yarn --frozen-lockfile " failed 3 times.
The command "yarn --frozen-lockfile" failed and exited with 1 during .

what can i do about it?

rossknudsen:
This is the error:

error semver@7.3.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"

I think you need to change the node version to 10 in the Travis config. I think I had to do that in my PR.

firsttris:
thx @rossknudsen that worked.

just coveralls is complaining now.

README.md
```diff
@@ -10,7 +10,24 @@ The engine that allows editors to build on top of Jest.
 
 This is only useful if you are interested in building an editor integration for Jest.
 
-For now as an end user, we'd recommend looking at either [vscode-jest](https://github.com/jest-community/vscode-jest/) or [majestic](https://github.com/Raathigesh/majestic/).
+For now as an end user, we'd recommend looking at either [vscode-jest](https://github.com/jest-community/vscode-jest/), [vscode-jest-runner](https://github.com/firsttris/vscode-jest-runner) or [majestic](https://github.com/Raathigesh/majestic/).
```
connectdotz: Not sure why `majestic` is here, it is not using this package isn't it? Actually I think I am fine to get rid of this whole sentence altogether.

README.md
```diff
@@ -10,7 +10,24 @@ The engine that allows editors to build on top of Jest.
...
+)
+```
+Parse is a static Jest parser which uses Babel 7 and supports js,jsx,mjs,ts,tsx files.   
+
+[Supported ECMAScript proposals](https://github.com/babel/babel/blob/928b9f8c9518284eac6d0598633f2ec373fc6d0c/packages/babel-parser/typings/babel-parser.d.ts#L97)
+
+- filePath = Path to the file you want to parse.
+- serializedData = Serialized data, will be used instead of the filePath if available (optional).
+- strictMode = If this option is activated the parser throws an exception if the filetype is not detected, defaults to false.
+
```
connectdotz, line 28: great way to start documenting the API 👍 , can we throw in an example, I am a big fan of "learn-by-example'

package.json
```diff
@@ -23,7 +23,7 @@
     "prepublish": "yarn run build",
     "test": "jest",
     "flow": "flow",
-    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(js|ts)\"",
+    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(js|ts)\" --fix",
```
connectdotz: let's split lint from lint-fix, if ci is going to use lint, I don't think it should be fixing anything...

src/parsers/babel_parser.js
```diff
@@ -171,3 +170,34 @@ export const parse = (file: string, data: ?string): ParseResult => {
...
+  'logicalAssignment',
+  'nullishCoalescingOperator',
+  'numericSeparator',
+  'objectRestSpread',
+  'optionalCatchBinding',
+  'optionalChaining',
+  'partialApplication',
+  'throwExpressions',
+  'topLevelAwait',
+];
```
connectdotz, line 198: looks like there is something missing from the [babel plugin list](https://github.com/babel/babel/blob/928b9f8c9518284eac6d0598633f2ec373fc6d0c/packages/babel-parser/typings/babel-parser.d.ts#L97), is it intentional?

connectdotz (commented):
@firsttris thanks for the update, we are almost there! a few non-blocking document comments, and a question about the plugin list.

Don't worry about coverage complaint, we should be good to go once we resolve the issues below

connectdotz:
hmm... on the 2nd thought, I think this is a good enough start, I can fix the remaining minor issues in another PR so we can free you to move on to the original thing you wanted to fix: 

> if you are happy with this PR, i would proceed update the jest parser to new features like test.each.

@firsttris if there is anything you want to add I can wait, otherwise I can merge it tonight.

README.md
firsttris: agree

README.md
firsttris: would be best to create an example with typescript and to link it in the readme (maybe follow up PR)

package.json
firsttris: that was actually not intentionally maybe that happen on merging Master

src/parsers/babel_parser.js
firsttris: there are conflicts between various plugins, tried to include all and removed the ones which caused conflicts

firsttris:
@connectdotz i'm rdy

connectdotz:
@firsttris thank you for your hard work 🙏   and sorry it took so long.

seanpoulter:
Yay! Thank you Tristan! 🥇

connectdotz:
@firsttris while testing the new typescript babel_parser, noticed a discrepancy vs the old typescript parser. While the old parser does not return the right "name" for the `jest.each`, it still returns the test block. The new parser completely skipped the block...

If you already started to work on the `jest.each` then I will just wait, otherwise, I can take a look to at least closing the parity...

firsttris:
Have not yet started looking in to this topic.

connectdotz:
ok, I did some quick fix at #47, would love to have you review it.

PR #32, opened 2020-01-25
Typescript conversion

rossknudsen:
This is a conversion to Typescript, removing Flow and updating build scripts.

Unit tests should be passing, but I need to manually test.  Maybe we can get a pre-release, or is there another simple way to substitute an npm package in my extension?


rossknudsen:
Hmmm, looks like I need to fix the linting...

rossknudsen:
Ah looks like the version of node on the CI server is too low to run one of the new dependencies.  Can we bump the version?

connectdotz:
while I like typescript a lot personally, I am not the only contributor in this repo, @stephtr @seanpoulter, are you guys ok to move the repo to typescript?

stephtr:
> while I like typescript a lot personally, I am not the only contributor in this repo, @stephtr @seanpoulter, are you guys ok to move the repo to typescript?

Personally, I definitely prefer working with TypeScript. It would also better fit to `vscode-jest`.

.eslintrc.js
```diff
@@ -5,20 +5,31 @@ module.exports = {
     browser: true,
     jest: true,
   },
-  extends: ['airbnb-base', 'plugin:flowtype/recommended', 'prettier', 'plugin:prettier/recommended'],
+  extends: [
+    'airbnb-typescript/base',
+    "airbnb/hooks",
```
stephtr: I guess we should use quotes consistently 😉

fixtures/dangerjs/github.example
```diff
@@ -1,4 +1,4 @@
-// @flow
+
```
stephtr, line 1: Why do some files start with an empty line?

fixtures/parser_tests.ts
```diff
@@ -5,18 +5,20 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import { ParseResult, NamedBlock, ItBlock, DescribeBlock } from "../src/parsers/parser_nodes";
+
 const fixtures = __dirname;
 
-function parserTests(parse: (file: string) => ParserResult) {
+function parserTests(parse: (file: string, data?: string | undefined) => ParseResult) {
```
stephtr: The `?` already contains `undefined`.

fixtures/parser_tests.ts
```diff
@@ -5,18 +5,20 @@
...
     block,
     sl: number,
     sc: number,
     el: number,
     ec: number,
-    name: ?string = null,
+    name: string | null = null,
   ) => assertBlock(block, {column: sc, line: sl}, {column: ec, line: el}, name);
-  const assertBlock = (block, start, end, name: ?string = null) => {
+  const assertBlock = (block, start, end, name: string | null = null) => {
```
stephtr: We could change this to just `name?: string`.

fixtures/parser_tests.ts
```diff
@@ -184,24 +186,24 @@ function parserTests(parse: (file: string) => ParserResult) {
     it('finds test blocks within describe blocks', () => {
       const data = parse(`${fixtures}/dangerjs/travis-ci.example`);
       const descBlock = data.describeBlocks[1];
-      expect(descBlock.children.length).toBe(4);
+      expect(descBlock.children?.length).toBe(4);
 
       // check test blocks, including the template literal
-      const found = descBlock.children.filter(
+      const found = descBlock?.children?.map(b => b as ItBlock)?.filter(
```
stephtr: In my opinion we don't need a `?.` after `map`.

fixtures/parser_tests.ts
```diff
@@ -414,6 +416,6 @@ function parserTests(parse: (file: string) => ParserResult) {
   });
 }
 
-module.exports = {
+export {
   parserTests,
-};
+}
```
stephtr: Why not use `export function parserTests`?

fixtures/parser_tests.ts
```diff
@@ -184,24 +186,24 @@ function parserTests(parse: (file: string) => ParserResult) {
     it('finds test blocks within describe blocks', () => {
       const data = parse(`${fixtures}/dangerjs/travis-ci.example`);
       const descBlock = data.describeBlocks[1];
-      expect(descBlock.children.length).toBe(4);
+      expect(descBlock.children?.length).toBe(4);
```
stephtr: Can `children` really be not an (empty) array? I'm just wondering about the change between old and new code.

src/Runner.ts
```diff
@@ -4,34 +4,33 @@
...
+
 import ProjectWorkspace from './project_workspace';
 import {createProcess} from './Process';
 
 // This class represents the running process, and
 // passes out events when it understands what data is being
 // pass sent out of the process
 export default class Runner extends EventEmitter {
-  debugprocess: ChildProcess;
+  debugprocess: ChildProcess | null = null;
```
stephtr: What about `debugprocess?: ChildProcess;`?

src/Runner.ts
```diff
@@ -85,6 +84,11 @@ export default class Runner extends EventEmitter {
     }
 
     this.debugprocess = this._createProcess(this.workspace, args);
+
+    if (!(this.debugprocess.stdout && this.debugprocess.stderr)) {
```
stephtr: For readability I would prefer `if(!A || !B)`.

src/__tests__/process.test.ts
```diff
@@ -30,24 +29,24 @@ describe('createProcess', () => {
     const args = [];
     createProcess(workspace, args);
 
-    expect(spawn.mock.calls[0][0]).toBe('jest');
-    expect(spawn.mock.calls[0][1]).toEqual([]);
+    expect((spawn as any).mock.calls[0][0]).toBe('jest');
```
stephtr: One could also use the `as Mock<any>` type from Jest.

src/parsers/parser_nodes.ts
```diff
@@ -26,25 +27,30 @@ export class ParsedRange {
...
 };
 
-export type ParsedNodeType = $Keys<typeof ParsedNodeTypes>;
+export type ParsedNodeType = keyof typeof ParsedNodeTypes;
 
 export class ParsedNode {
   type: ParsedNodeType;
 
-  start: Location;
+  start: Location | undefined = undefined;
```
stephtr: On such occasions `start?: Location;` would be sufficient.

src/parsers/parser_nodes.ts
```diff
@@ -1,14 +1,15 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable max-classes-per-file */
```
stephtr: In my opinion we could also get rid of the `max-classes-per-file` policy project-wide.

src/parsers/parser_nodes.ts
```diff
@@ -26,25 +27,30 @@ export class ParsedRange {
...
+  describe: 'describe';
+  expect: 'expect';
+  it: 'it';
+  root: 'root';
+} = {
   describe: 'describe',
   expect: 'expect',
   it: 'it',
   root: 'root',
 };
```
stephtr, line 38: It has to be tested, but this could be replaced by either an enum or `export type ParsedNodeTypes = 'describe' | 'expect' | ...`.

src/parsers/parser_nodes.ts
```diff
@@ -26,25 +27,30 @@ export class ParsedRange {
...
-  start: Location;
+  start: Location | undefined = undefined;
 
-  end: Location;
+  end: Location | undefined = undefined;
 
   file: string;
 
-  children: ?Array<ParsedNode>;
+  children?: Array<ParsedNode>;
```
stephtr: `ParsedNode[]` is easier to read than `Array<...>`.

stephtr (commented):
I didn't have time to go through all changes, but these are my comments for some of them.

.eslintrc.js
rossknudsen: My choice would have been for double quotes, but I tried to stick with the Prettier rules that were already present.

fixtures/dangerjs/github.example
rossknudsen: Thats a good question.  It took me ages to figure out why some unit tests were failing.  Turns out the line numbers are hard-coded in them and I was trying to change the minimal amount of code, so that this PR was readable.

So I didn't update the tests with new line numbers.

rossknudsen:
Huh, somehow the coverage dropped...

src/parsers/parser_nodes.ts
rossknudsen: I agree.  Pretty sure this is a Prettier/ESLint rule.

src/parsers/parser_nodes.ts
rossknudsen: Agreed.  I was tempted to rewrite some of the parser code relying on TS more, but I see that #30 is already rewriting the parser code.  It would probably make more sense to merge those changes first.

.eslintrc.js
stephtr: I was just refering to some `extends` strings being single- and some being doublequoted. I would prefer enforcing consistent quotes for the whole project.

src/parsers/parser_nodes.ts
rossknudsen: Actually having read that PR, I'm inclined to wait until it is merged because it will remove the typescript parser. That reduces the delta in this PR.

rossknudsen:
> while I like typescript a lot personally, I am not the only contributor in this repo, @stephtr @seanpoulter, are you guys ok to move the repo to typescript?

Fair enough, I was going by this [comment](https://github.com/jest-community/jest-editor-support/blob/master/src/Settings.js#L16-L20) which is probably old/stale now.  I ran across the comment because I wanted to access more of the Jest settings.  I guess I went down a bit of a rabbit hole with this one...

src/parsers/parser_nodes.ts
rossknudsen: Sure.  

What is your opinion of placement of the copyright notice vs linting configuration?  I had just left it where the intellisense put it, at the very top.

.eslintrc.js
rossknudsen: Ah my bad.  Was reading that comment on mobile.  Not sure why I've got a mixture there.  Will fix.

src/__tests__/process.test.ts
rossknudsen: Cool, slightly more type-safe.  Any ideas on replacing the `any` part of the type with something more specific?

fixtures/parser_tests.ts
seanpoulter: 👍 

How did you add these @rossknudsen?

seanpoulter:
Thanks for the huge effort @rossknudsen! 👏 

--

> [@rossknudsen wrote](https://github.com/jest-community/jest-editor-support/pull/32#issue-367154265)
>
> Unit tests should be passing, but I need to manually test. Maybe we can get a pre-release, or is there another simple way to substitute an npm package in my extension?

Yes, you can substitute npm packages using "linking". Here's the docs for `npm` and `yarn` that are probably more clear than what I'd write. ;)
* https://docs.npmjs.com/cli/link.html
* https://legacy.yarnpkg.com/lang/en/docs/cli/link/

--

> [@connectdotz wrote](https://github.com/jest-community/jest-editor-support/pull/32#issuecomment-578521269):
>
> while I like typescript a lot personally, I am not the only contributor in this repo, @stephtr @seanpoulter, are you guys ok to move the repo to typescript?

I've got two answers for this: one with best interest of the project in mind, and the other from my own perspective.

<ol><li>With the best interest in the project in mind, I'd say yes, switching to TypeScript makes sense because folks are more familiar with TypeScript than Flow. I don't think any of the other reasons I've considered sway me one way or the other.
  <table>
  <tr>
    <th>Pros</th>
    <td><ul>
      <li>More type information will help folks use the library in lieu of documentation</li>
      <li>We can leverage more of the types exported from the facebook/jest repo</li>
      <li>TypeScript has more adoption than Flow [1] [2]</li>
      <li>TypeScript may have better IDE integration than Flow</li>
      <li>TypeScript might improve our code quality</li>
    </ul></td>
  </tr>
  <tr>
    <th>Cons</th>
    <td><ul>
      <li>TypeScript is still a barrier to entry over "regular" "JavaScript" -- not everyone will keep up with the ES20xx language features</li>
      <li>Spending time on test case coverage and code review may be a better time investment [3]</li>
    </ul></td>
  </tr>
</table></li>

<li>With my own interest in mind, I see this is a PR that's too big for me to approach so I'm biased towards keeping the status quo. While I think this is adding a lot of value, we may have missed some of the conversations we need like:
<ul>
<li>Do the maintainers have the capacity to review the PR? (I don't)</li>
<li>Can we make the switch from Flow to TypeScript in incremental, easy to review PRs?</li>
<li>Do we need to discussion what TS config we'll use, and is it familiar? (As an arbitrary example, this is the first time I've seen optional chaining in a repo and sometimes they feel icky, like in tests).</li>
<li>What test cases do we need to have confidence in this, especially if it's too big to review?</li>
<li><strong>(edited to add this)</strong>: Are we OK to migrate AND refactor at the same time? For example, can we convert the Flow Maybe to the TypeScript optional?</li>
</ul>
</ol>


\[1]: https://2019.stateofjs.com/javascript-flavors/typescript/
\[2]: https://2019.stateofjs.com/javascript-flavors/other-tools/
\[3]: https://medium.com/javascript-scene/the-typescript-tax-132ff4cb175b

.eslintrc.js
```diff
@@ -5,20 +5,36 @@ module.exports = {
     browser: true,
     jest: true,
   },
-  extends: ['airbnb-base', 'plugin:flowtype/recommended', 'prettier', 'plugin:prettier/recommended'],
+  extends: [
+    'airbnb-typescript/base',
+    'airbnb/hooks',
```
seanpoulter: Did you sneak React hooks into the project too? 😄 
```suggestion
```

fixtures/parser_tests.ts
rossknudsen: @seanpoulter if you mean this specific piece of code, it now looks like this:

```ts
// check test blocks, including the template literal
const found = descBlock.children
.map(b => b as NamedBlock)
.filter(
    b =>
    b.name === 'needs to have a PR number' ||
    b.name === 'does not validate without josh' ||
    b.name === 'does not validate when ${key} is missing'
);
expect(found.length).toBe(3);
```

I was just trying to assert that the type of the array elements were `NamedBlock`s (originally `ItBlock`s).  There is potentially another way to write this which is arguably more elegant.  Now that the `children` property is not potentially null, it is a lot clearer.

If you are talking more generally, you might need to give me some more context.

.eslintrc.js
rossknudsen: Lol, force of habit!  I'll remove.

src/test_reconciler.ts
```diff
@@ -72,16 +71,16 @@ export default class TestReconciler {
     return assertions.map(assertion => {
       // Failure messages seems to always be an array of one item
       const message = assertion.failureMessages && assertion.failureMessages[0];
-      let short = null;
-      let terse = null;
-      let line = null;
+      let short: string | undefined;
+      let terse: string | undefined;
+      let line: number | undefined;
```
seanpoulter, line 76: The change from `null` to `undefined` stands out to me as a potentially risky API change. The risk is that a caller uses all three types from the Flow Maybe type (`type | null | void (undefined)`) to differentiate between values that aren't in the map/object, are invalid, or are a valid number. It makes me worry that I'd/we'd have to do a more thorough code review or testing.

In this particular case, the caller seems fine but there's still the rough edge of how we're using `this.lineOfError(...) || undefined` to do ... something. What would ever be falsy enough to end up as `undefined`?! I actually had to run this in the Dev Tools console just to check the falsy precedence:

image

I guess I learned (or relearned) that. ☺️  Since I'll probably forget that the next time I look at it, I'd probably suggest going for:

```diff
-         line = this.lineOfError(message, filename) || undefined;
+         line = this.lineOfError(message, filename);
```
with `lineOfError` returning `undefined`. I'm not a fan of the magic numbers like `[1]` in there so I'd hope to see something like:
```diff
-   lineOfError(message: string, filePath: string): number | null {
-     const filename = path.basename(filePath);
-     const restOfTrace = message.split(filename, 2)[1];
-     return restOfTrace ? parseInt(restOfTrace.split(':')[1], 10) : null;
-   }
+   lineOfError(message: string, filepath: string): ?number {
+     const filename = path.basename(filepath);
+     const [_msgIncludingFilename, msgAfterFilename] = message.split(filename, 2);
+     if (! msgAfterFilename) return;
+     const [_colon, lineNumber] = msgAfterFilename(':');
+     return parseInt(lineNumber, 10);
+   }
```
--

The underlying question for the folks reviewing is are we OK with migrating **and** refactoring at the same time. Here's the docs for the [Flow Maybe type](https://flow.org/en/docs/types/maybe/).

rossknudsen:
@seanpoulter I appreciate the time taken to write your thoughts regarding this PR.  I reviewed the changed files again and it is substantial.  My original intent was to avoid a massive delta as it has potential to introduce regressions etc and part of the reason why people wouldn't want to convert in the first place.  Also 100% appreciate your limited time in reviewing this.

In light of all this, I stepped back and thought about whether its possible to take an incremental approach to conversion.  So I looked at the recommendations for conversion from Microsoft and cherry-picked some changes from this branch into a new branch.  If you like this approach I can open another PR for review.  I think I've kept everything operational while converting just one file to Typescript.

connectdotz:
> In light of all this, I stepped back and thought about whether its possible to take an incremental approach to conversion... If you like this approach I can open another PR for review. I think I've kept everything operational while converting just one file to Typescript.

I like that! 👍

connectdotz:
superseded by #33

PR #33, opened 2020-01-27
Typescript-lite - minimal conversion to Typescript

rossknudsen:
This is another attempt at converting to TS.  This time, a minimal approach was taken since the previous attempt was too large to review.  Since Babel can translate JS with Flow as well as Typescript, this works well (pretty sure the TS compiler doesn't understand Flow).  So the changes are mostly to configuration files with a single example JS file converted to TS as a proof of concept (/src/index.js => /src/index.ts).

Inclusion of a `.gitattributes` file and changing single quotes to double quotes in the `package.json` scripts was introduced to help with Windows compatibility.


tsconfig.json
```diff
@@ -0,0 +1,64 @@
+{
+    "compilerOptions": {
+      /* Basic Options */
+      "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
```
seanpoulter: Any reason to choose `es5`? Is the Node v6 in the TravisCI config the limiting factor?

I've been out of the loop for a while and forget where to check Node support for those target versions. I used to use https://node.green to check this kinda thing but it seems like that's a subset of these tables which includes https://kangax.github.io/compat-table/es6/

package.json
```diff
@@ -18,21 +18,25 @@
...
   "scripts": {
-    "build": "babel src/ -d build/",
+    "build": "babel src/ -d build/ --extensions \".js,.ts\" --source-maps inline",
+    "build:types": "tsc --emitDeclarationOnly",
     "prepublish": "yarn run build",
     "test": "jest",
     "flow": "flow",
-    "lint": "eslint '?(__mocks__|src|tests)/**/*.js'",
+    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(ts|js)\"",
     "ci": "yarn flow && yarn lint && yarn test --coverage",
```
stephtr: Does `yarn type-check` currently fail? Otherwise I would add it to `ci`.

rossknudsen: I'll check. It might because I added the allowJs flag in tsconfig.json. But if it works without that flag I'll definitely add it.

stephtr (approved):
To me it looks good.

.eslintrc.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
   },
   plugins: ['flowtype', 'prettier'],
   rules: {
-    'prettier/prettier': 'error',
+    'prettier/prettier': 'off',  // disabled temporarily to prevent un-necessary delta in conversion to TS.
```
connectdotz: My vote is to keep this unchanged ('error'), as long as we keep our change progressive, I'd rather live with the whitespace diff (which we could probably turn off during diffing) than inconsistent coding style.

I didn't see any typescript specific support (@typescript-eslint/xxx). Should we add those? Not sure if it can live in the same config with "flow", worst case separate config?

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "typescript.validate.enable": false,
-  "javascript.validate.enable": false,
+  "typescript.validate.enable": true,
+  "javascript.validate.enable": true,
```
connectdotz: with `"javascript.validate.enable": true`, I started to see ts error in the non-converted js files. Should we follow [this](https://github.com/flowtype/flow-for-vscode#setup) for the flow js files?

babel.config.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
       { useBuiltIns: 'usage', corejs: '3' }
     ],
     '@babel/flow',
+    '@babel/typescript'
```
connectdotz, line 8: do we want to rely on babel to transpile typescript files or just have typescript (tsc) do both transpile and type checking and generation? Furthermore, I am wondering maybe we will not need babel once flow dependency is removed... therefore, maybe not add any new dependency where tsc can handle? 🤔

package.json
```diff
@@ -18,21 +18,25 @@
     ]
   },
   "scripts": {
-    "build": "babel src/ -d build/",
+    "build": "babel src/ -d build/ --extensions \".js,.ts\" --source-maps inline",
+    "build:types": "tsc --emitDeclarationOnly",
```
connectdotz, line 22: if we agree to use babel for flow type only, then this should be adjusted accordingly.

package.json
```diff
@@ -18,21 +18,25 @@
...
     "prepublish": "yarn run build",
```
connectdotz, line 23: "prepublish" should call both build js and ts and types.

package.json
connectdotz: we should just add `yarn prepublish` or `yarn build-all` to make sure the code does build, which includes type-check.

tsconfig.json
```diff
@@ -0,0 +1,64 @@
...
+      // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
+      // "removeComments": true,                /* Do not emit comments to output. */
+      // "noEmit": true,                        /* Do not emit outputs. */
+      // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+      // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+      // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+  
+      /* Strict Type-Checking Options */
+      "strict": false,                          /* Enable all strict type-checking options. */
+      "noImplicitAny": false,                   /* Raise error on expressions and declarations with an implied 'any' type. */
```
connectdotz, line 27: I know why we have to have this set to true now due to many files haven't been migrated over. Please make a note that this should be reset to false once the migration is completed.

tsconfig.json
```diff
@@ -0,0 +1,64 @@
...
+      // "incremental": true,                   /* Enable incremental compilation */
+      // "tsBuildInfoFile": "./",               /* Specify file to store incremental compilation information */
+      // "removeComments": true,                /* Do not emit comments to output. */
+      // "noEmit": true,                        /* Do not emit outputs. */
+      // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+      // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+      // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+  
+      /* Strict Type-Checking Options */
+      "strict": false,                          /* Enable all strict type-checking options. */
```
connectdotz, line 26: why not "true"?

connectdotz (commented):
thanks, @rossknudsen for adopting the incremental approach. I do see the potential complexity of living in flow and typescript worlds simultaneously, but I guess that is what we agree as a trade-off for the massive one-time conversion...

This incremental approach made it easier to get the typescript tooling right. I have added a few comments, let's discuss and see if we can establish a solid starting point.

.eslintrc.js
seanpoulter: > My vote is to keep this unchanged ('error'), as long as we keep our change progressive, I'd rather live with the whitespace diff (which we could probably turn off during diffing) than inconsistent coding style.

May I suggest that we keep it 'off' in this MR and follow-up with a "quick" whitespace-only MR that'll be an almost automatic approval?

<br/>

> I didn't see any typescript specific support (@typescript-eslint/xxx). Should we add those? Not sure if it can live in the same config with "flow", worst case separate config?

👍  It's a small repo. Should we go with the recommendations? I don't know how much these diverge from our coding style. If it's bad, maybe it should follow the whitespace MR. ;). https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md

.vscode/settings.json
seanpoulter: So, keep it off for JS files and use the linting/flow tools from the CLI until we're through this? 👍

babel.config.js
seanpoulter: I don't have a strong opinion either way. Once Flow is gone, I'm AOK with just `tsc`.

package.json
seanpoulter: You OK with chaining them?
```suggestion
     "prepublish": "yarn run build && yarn run build:types",
```

package.json
```diff
@@ -18,21 +18,25 @@
...
     "test": "jest",
     "flow": "flow",
-    "lint": "eslint '?(__mocks__|src|tests)/**/*.js'",
+    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(ts|js)\"",
```
seanpoulter: Non-blocking OCD comment - this one's different 🕵️‍♀️ :
```suggestion
    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(js|ts)\"",
```

.eslintrc.js
rossknudsen: Cool, happy to leave Prettier on.

Regarding @typescript-eslint/xxx linting rules/presets, I have been playing with the linting recently.  Hopefully I'll get that up shortly.

package.json
```diff
@@ -18,21 +18,25 @@
...
     "ci": "yarn flow && yarn lint && yarn test --coverage",
-    "prettier": "prettier --check '?(__mocks__|src|tests)/**/*.js'",
+    "prettier": "prettier --check \"?(__mocks__|src|tests)/**/*.+(js|ts)\"",
     "prettier-write": "prettier --write",
-    "prettier-project": "yarn prettier-write '?(__mocks__|src|tests)/**/*.js'"
+    "prettier-project": "yarn prettier-write \"?(__mocks__|src|tests)/**/*.+(js|ts)\"",
+    "type-check": "tsc --noEmit",
+    "type-check:watch": "npm run type-check -- --watch"
```
seanpoulter: Just a non-blocking comment:
> Oh JavaScript ... `yarn` this (prettier-write) ... `npm` that. 😅

tsconfig.json
seanpoulter: Want to make a GitHub project to track it, or should we put a TODO list on an issue? 😁

babel.config.js
rossknudsen: My preference is to use the TS compiler for TS projects rather than babel.  

However, while there is a mixture of JS with Flow types and TS code, I reckon that babel is a better choice.  That is because babel can understand TS, Flow and JS simultaneously, while the TS compiler doesn't understand Flow (if you know of a way to precompile the Flow types out or some other voodoo let me know).  I've played around converting a few other files and the difficulties soon become apparent.  I might push up a couple more code changes that expose the issues.

I think in the long term, moving to a configuration that is the same/similar to what Jest itself uses would be the ideal.

seanpoulter (approved):
Wow, thank you so much for splitting up your epic MR @rossknudsen! 🎉 It looks great. I'm sure we'll dial in the configuration as we go along. I don't think there's any of those config changes I'd consider blocking so I'll approve it for now. We can start to converge on the tsconfig from <your favourite project> as we go.

(What were you thinking of? I had checked facebook/jest, jest-community/vscode-jest, and @orta's (?) microsoft/TypeScript-Babel-Starter).

package.json
rossknudsen: I noticed that there is `index.d.ts` in the project root.  This seems to have type definitions for all the public types in the project and I didn't want to double up.  Maybe when the entire codebase is TS, then include the built definitions?  What do you think?

Also I can't see what steps are performed for a release - I presume the types are added manually?

tsconfig.json
rossknudsen: Sure, do we need an issue to track this?  As mentioned above, I think we should look at making this as close to Jest's configuration as feasible.  I don't think we need a comment in source, so much of the config will change once we get to 100% TS.

tsconfig.json
rossknudsen: I know right 😄

.vscode/settings.json
rossknudsen: The Flow extension recommends turning off the JS validation because it seems to use the TS compiler which then complains about Flow type syntax being invalid.  I'm happy leaving the JS validation off, but the TS validation seems to work Ok.

package.json
rossknudsen: See my thoughts on going with babel in the short term above.

package.json
stephtr: > Maybe when the entire codebase is TS, then include the built definitions? What do you think?

I prefer that, since we want to stick to our index.d.ts until the whole project and all types have been converted.

rossknudsen: > Also I can't see what steps are performed for a release - I presume the types are added manually?

Ah I see that the typings are referenced in the `package.json`, which is probably how that works.

package.json
```diff
@@ -18,31 +18,40 @@
...
+    "build": "babel src/ -d build/ --extensions \".js,.ts\" --source-maps inline",
+    "build:types": "tsc --emitDeclarationOnly",
     "prepublish": "yarn run build",
     "test": "jest",
     "flow": "flow",
-    "lint": "eslint '?(__mocks__|src|tests)/**/*.js'",
-    "ci": "yarn flow && yarn lint && yarn test --coverage",
-    "prettier": "prettier --check '?(__mocks__|src|tests)/**/*.js'",
+    "lint": "eslint \"?(__mocks__|src|tests)/**/*.+(js|ts)\"",
+    "ci": "yarn type-check && yarn lint && yarn test --coverage && yarn flow",
```
rossknudsen, line 27: Note that I changed the order of the 'ci' script.  That's because all the other checks pass, but the flow script fails.  It can't figure out how to import `project_workspace.ts`:

```
Error ---------------------------------------------------------- src/__tests__/runner.test.js:19:30

Cannot resolve module `../project_workspace`.

   19| import ProjectWorkspace from '../project_workspace';
                                    ^^^^^^^^^^^^^^^^^^^^^^
```
I've tried playing around with the `.flowconfig` file but I can't figure out how to get it to work.  It would be a shame if we couldn't maintain the flow type checking.

tsconfig.json
rossknudsen: I think when I created the config (`tsc --init`) that's what it was set to.  Happy to change it to your preference, do we know what runtimes that we're supporting?  We might be able to bump to node v8 now.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "6"
+  - "8"
```
rossknudsen, line 3: I bumped the node version because it was erroring out saying that **@typescript-eslint/eslint-plugin** required a later version.  If this is not acceptable, I can have a look at other options.

connectdotz: I think node 8 is acceptable, vscode-jest also supports node >= 8. @firsttris will this node version bump ok for your repo?

firsttris: yes

package.json
rossknudsen: Awesome, managed to figure it out and it passes!

tsconfig.json
rossknudsen: Updated this to ES6 since we've bumped up to node v8 which the link(s) you posted seem in indicate is ES6 compatible.

tsconfig.json
rossknudsen: Cool, I added #34 to track the conversion.  If you'd like to upgrade it to a project feel free.

rossknudsen:
Hey all.  Just wanted to check if there was anything missing from this PR?  I'm happy with it and as far as I can tell, I've either updated the code in response to comments or provided justification for the current changes.  But I'm not sure what the etiquette is and this is my first contribution to OSS that has peer review.

Not being pushy, just wanting to check that I'm not holding things up 😃

connectdotz:
> Just wanted to check if there was anything missing from this PR?

@rossknudsen thanks for being patient, I checked the latest change and tested with `vscode-jest`, it all seemed fine! 👍  I think this PR will kickoff the ts migration that we all wanted for a while 😄 

I will merge it later today.

connectdotz (approved):

PR #35, opened 2020-02-09
Converted core files to Typescript

rossknudsen:
Next PR for TS conversion.  This time all files except tests and parsers - which will involve the most change.

Note that the amount of change to `TestResult.js|ts` shows the diff as a deletion and an addition.  Making it slightly harder to see what has changed.  You may want to manually diff them (unless there is a way to get git to display them as a rename/modification).


connectdotz:
@seanpoulter or @stephtr do you guys have some time to look at this? I am quite swamped recently, feel bad that this PR has been sitting idle for a while...

rossknudsen:
If you feel that @firsttris's PR is close, then happy to wait and I'll rebase before review.

src/Settings.ts
```diff
@@ -20,16 +19,16 @@ type Glob = string;
 // in favor of `@jest/types`, which exports the full config interface.
 
 type ProjectConfiguration = {
```
stephtr, line 21: In principle we could also use interfaces, but since there's hardly any difference, it's probably better to stick to the already existing types.

types/TestResult.ts
```diff
@@ -0,0 +1,228 @@
+/**
+ * Copyright (c) 2014-present, Facebook, Inc. All rights reserved.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
```
stephtr: We don't need that anymore.

types/TestResult.ts
```diff
@@ -0,0 +1,228 @@
...
+};
+
+export type Suite = {
+  title: string;
+  suites: Suite[];
+  tests: AssertionResult[];
+};
+
+export type TestResult = {
+  // console: ?ConsoleBuffer,
```
stephtr: If we keep that comment, I'd also transform that to TS.

types/TestResult.ts
```diff
@@ -0,0 +1,228 @@
...
+  merge: (data: Record<string, any>) => void;
+  getCoverageSummary: () => FileCoverage;
+  data: RawCoverage;
+  addFileCoverage: (fileCoverage: RawFileCoverage) => void;
+  files: () => string[];
+  fileCoverageFor: (file: string) => FileCoverage;
+};
+
+export type SerializableError = {
+  code?: unknown;
```
stephtr, line 64: Is there a reason for using `unknown` here but `any` everywhere else?

types/TestResult.ts
```diff
@@ -0,0 +1,228 @@
...
+export type RawFileCoverage = {
+  path: string;
+  s: {[statementId: number]: number};
+  b: {[branchId: number]: number};
+  f: {[functionId: number]: number};
+  l: {[lineId: number]: number};
+  fnMap: {[functionId: number]: any};
+  statementMap: {[statementId: number]: any};
+  branchMap: {[branchId: number]: any};
+  inputSourceMap?: Record<string, any>;
```
stephtr: I wouldn't use `Record` with a string as first template parameter, especially since all properties before are dicts.

stephtr (commented):
Sorry, lately my spare time is very limited.
I didn't notice any larger issues, just a few comments.

src/Snapshot.ts
```diff
@@ -5,44 +5,43 @@
...
 import {getASTfor} from './parsers/babylon_parser';
 
 type Node = any;
 
 type SnapshotMetadata = {
-  exists: true | false,
-  name: string,
-  node: Node,
-  content?: string,
+  exists: true | false;
```
stephtr:
```suggestion
  exists: boolean;
```
😉

src/test_reconciler.ts
```diff
@@ -5,13 +5,12 @@
...
 
 import path from 'path';
-import type {TestFileAssertionStatus, TestAssertionStatus, TestReconciliationState, Location} from './types';
+import {TestFileAssertionStatus, TestAssertionStatus, TestReconciliationState, Location} from './types';
 
-import type {FormattedAssertionResult, FormattedTestResults} from '../types/TestResult';
+import {FormattedAssertionResult, FormattedTestResults} from '../types/TestResult';
 
 /**
  *  You have a Jest test runner watching for changes, and you have
```
stephtr, line 16: Can we get rid of the following constructor by writing `fileStatuses: {[key: string]: TestFileAssertionStatus} = {};`?

```ts
export default class TestReconciler {
  fileStatuses: {[key: string]: TestFileAssertionStatus};
  constructor() {
    this.fileStatuses = {};
  }
```

types/TestResult.ts
rossknudsen: I was following a conversion guide.  Interestingly, this type seems to exist in the [Jest repo](https://github.com/facebook/jest/blob/master/packages/jest-test-result/src/types.ts#L20-L25) and they have defined it as `unknown` also.  I'm not particularly fussed about changing it to `any`, so if you think that's better, I'll change it.

I am interested in the fact that the types are defined in Jest however...

stephtr: Since they use `unknown` I would also stick to that.

connectdotz:
@rossknudsen we are moving toward a major release, so I wonder maybe we could include this typescript migration if you are up for it?

rossknudsen:
@connectdotz I'm keen as to get this in the next release.  I've just updated this PR to include the latest changes.  Since this PR has been reviewed quite a bit already, I won't update the parsers.  We can review that in a separate PR.

connectdotz:
I know we talked you out of doing a full conversion, but after my experience with the last PR, which is quite painful in a mixed world, I have a change of heart that I think we are better off to take the pain and convert it all the way to typescript and modem eslint. 

If others are busy I can work with you on that but we need to have a **no-logic change conversion**. If anything that is not ideal, just mark it with a "TODO" comment and we will deal with them in subsequent PRs.

I am ok with removing the `types/xxx` folder that seems to for the flow types support. Let's use typescript for compile and type generation instead of babel.

As far as eslint, can you take a look at `vscode-jest` where we just went through a recent eslint migration and I think the rules are clean and compliant with the recommended sets.

rossknudsen:
Sweet - I agree with your proposal.  I've updated most of the remaining files.  Need some work around the linting and compiler setup.  Not sure what your thoughts were on having a common branch.

Granted you permission to my repository - but if you have a better idea, let me know.  In hindsight, I guess you can add a remote pointing at my repo...  Oh well, whichever way you want to work.

connectdotz:
@rossknudsen as long as you have the permission to allow maintainers to edit your PR, I should be able to submit changes. Feel free to just submit your changes, when you are ready I can clone this branch and test it locally.

src/__tests__/snapshot.test.ts
```diff
@@ -5,13 +5,12 @@
...
  * of patent rights can be found in the PATENTS file in the same directory.
  *
- * @flow
  */
 
 import path from 'path';
 import Snapshot from '../Snapshot';
 
-const snapshotHelper = new Snapshot();
+const snapshotHelper = new Snapshot(undefined);
```
stephtr: Wouldn't it be better to change [Snapshot.ts#L87](https://github.com/jest-community/jest-editor-support/blob/5c294af402a75d8c2e1240380298229c7bc2942d/src/Snapshot.ts#L87) to `parser?: any` instead?

src/parsers/parser_nodes.ts
```diff
@@ -24,17 +23,15 @@ export class ParsedRange {
...
-// export type ParsedNodeType = 'expect' | 'describe' | 'it' | 'ROOT';
+export type ParsedNodeType = 'expect' | 'describe' | 'it' | 'root';
 
-export const ParsedNodeTypes = {
+export const ParsedNodeTypes: Record<ParsedNodeType, ParsedNodeType> = {
   describe: 'describe',
   expect: 'expect',
   it: 'it',
   root: 'root',
 };
```
stephtr: Couldn't one replace that by an enum? That way the Record would be generated automatically.
For example:
```ts
export enum ParsedNodeTypes {
  describe = 'describe',
  expect = 'expect',
  it = 'it',
  root = 'root',
};

type ParsedNodeType = keyof typeof ParsedNodeTypes;
```

stephtr (commented):
I only had time for a short glimpse, attached are two small details I noticed.

src/__tests__/snapshot.test.ts
rossknudsen: Cool, done.

src/parsers/parser_nodes.ts
rossknudsen: Also done.

seanpoulter:
I apologize. This has been open for a while. That leads to a few questions:

* Are we still interested in migrating to TS?
* Do you think it's worth resolving conflicts or starting over?
* If we are starting over, can we work in **tiny** steps (like Array<T> => T[]) so we're not held up in review?

connectdotz:
@seanpoulter, welcome back!😃

Yes, I think we are still interested in moving this PR to TS. Especially with the upcoming Jest v30, we should upgrade our snapshot and other packages anyway. 

Let's start afresh, as this PR is too outdated. I am okay with either a progressive change or one big migration PR. As we have seen, progressive migration creates its own complications. I now have second thoughts about whether it's worth it, given that it is supposed to be for the transition period only.

@rossknudsen, thanks for your effort, and sorry we didn't finish it. I will close this PR for now and we shall consider the conversion in Jest v30 upgrade, if not sooner.

PR #39, opened 2020-04-20
address orphan process issue

connectdotz:
see jest-community/vscode-jest#550, specifically the 2nd scenario mentioned [here](https://github.com/jest-community/vscode-jest/issues/550#issuecomment-615930298)

> when the spawned process spawns another child process in the background, such as in react-scripts use case, even when we close the jest process, the react-script spawned process will not be terminated. 

It is clearly mentioned in nodejs [child_process kill](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal), I think we just missed it when switching to the `shell` option 😱 ...

> On Linux, child processes of child processes will not be terminated when attempting to kill their parent. This is likely to happen when running a new process in a shell or with the use of the shell option of ChildProcess

anyway, there are a few possible ways to fix this issue, looks like on windows, which we have been using shell option way back, it spawned an external `taskkill` to kill all child processes. At least on Unix like systems, it seems we could utilize the system's "process group" mechanism to create and terminate the child processes without spawning another external command nor introducing a 3rd party library... 

The "process group" mechanism in nodejs means to spawn a process with the "detached" option, which created a new process group that we can use later to kill every process within by `process.kill()` with the negative PID (served as PGID)

I tested this on mac and it seems to work as expected. I am a bit concern if this will work for all systems, so still keep the current behavior as the fallback. 

Also not sure if the `detached` flag has any unexpected effect on windows and if we should only turn this on for non-windows?

@stephtr or anyone has a windows system, would appreciate some quick test to verify all is well there.

### DIFF
--- a/src/__tests__/process.test.js
+++ b/src/__tests__/process.test.js
@@ -82,7 +82,7 @@ describe('createProcess', () => {
     expect(spawn.mock.calls[0][2].cwd).toBe(workspace.rootPath);
   });
 
-  it('should set the "shell" and "detached" property', () => {
+  it('should set the "shell" property', () => {
     const expected = true;
     const workspace: any = {pathToJest: ''};
     const args = [];
@@ -90,5 +90,23 @@ describe('createProcess', () => {
 
     expect(spawn.mock.calls[0][2].shell).toBe(expected);
     expect(spawn.mock.calls[0][2].detached).toBe(expected);
+  });
+  it('should set "detached" to true for non-windows system', () => {
+    const workspace: any = {pathToJest: ''};
+    const args = [];
+
+    const savedProcess = process;
+    const processMock = {...process};
+    global.process = processMock;
+
+    processMock.platform = 'darwin';
+    createProcess(workspace, args);
+    expect(spawn.mock.calls[0][2].detached).toBe(true);
+
+    processMock.platform = 'win32';
+    createProcess(workspace, args);
+    expect(spawn.mock.calls[1][2].detached).toBe(false);
+
+    global.process = savedProcess;
   });
 });
--- a/src/__tests__/process.test.js
+++ b/src/__tests__/process.test.js
@@ -89,7 +89,6 @@ describe('createProcess', () => {
     createProcess(workspace, args);
 
     expect(spawn.mock.calls[0][2].shell).toBe(expected);
-    expect(spawn.mock.calls[0][2].detached).toBe(expected);
   });
   it('should set "detached" to true for non-windows system', () => {
     const workspace: any = {pathToJest: ''};
--- a/src/__tests__/process.test.js
+++ b/src/__tests__/process.test.js
@@ -82,12 +82,13 @@ describe('createProcess', () => {
     expect(spawn.mock.calls[0][2].cwd).toBe(workspace.rootPath);
   });
 
-  it('should set the "shell" property', () => {
+  it('should set the "shell" and "detached" property', () => {
     const expected = true;
     const workspace: any = {pathToJest: ''};
     const args = [];
     createProcess(workspace, args);
 
     expect(spawn.mock.calls[0][2].shell).toBe(expected);
+    expect(spawn.mock.calls[0][2].detached).toBe(expected);
   });
 });
--- a/src/__tests__/runner.test.js
+++ b/src/__tests__/runner.test.js
@@ -44,6 +44,17 @@ const _path = jest.requireActual('path');
 const fixtures = _path.resolve(__dirname, '../../fixtures');
 
 describe('Runner', () => {
+  const processKill = jest.fn();
+  let originalProcessKill;
+
+  beforeAll(() => {
+    originalProcessKill = process.kill;
+    (process: any).kill = processKill;
+  });
+
+  afterAll(() => {
+    (process: any).kill = originalProcessKill;
+  });
   describe('constructor', () => {
     it('does not set watchMode', () => {
       const workspace: any = {};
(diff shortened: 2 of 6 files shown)


stephtr:
On Windows this unfortunately spawns a couple of cmd windows ("On Windows [...] the child will have its own console window.").

Since, as far as I've tested, [vscode-jest#550](https://github.com/jest-community/vscode-jest/issues/550) doesn't affect Windows, I would enable `detached` just on non-Windows systems.

connectdotz:
thanks, @stephtr. I made the change you suggested, please give it a try when you get a chance...

src/__tests__/process.test.js
```diff
@@ -91,4 +91,22 @@ describe('createProcess', () => {
     expect(spawn.mock.calls[0][2].shell).toBe(expected);
     expect(spawn.mock.calls[0][2].detached).toBe(expected);
```
stephtr: That line has to be removed (fails on Windows).

stephtr:
Besides the failing test, it seems fine to me 👍

stephtr (approved):

stephtr:
I have only tested it on Windows, but to me the PR looks fine. Maybe someone should still test it on Linux, just to be sure.

src/__tests__/process.test.js
```diff
@@ -90,4 +90,22 @@ describe('createProcess', () => {
...
+    processMock.platform = 'darwin';
+    createProcess(workspace, args);
+    expect(spawn.mock.calls[0][2].detached).toBe(true);
+
+    processMock.platform = 'win32';
+    createProcess(workspace, args);
+    expect(spawn.mock.calls[1][2].detached).toBe(false);
+
+    global.process = savedProcess;
+  });
```
seanpoulter, line 110: Non-blocking: It looks like a test failure could leave us with the incorrect `process.platform`. That said, I'd rather you leave it as is and focus on cutting the release. We can clean it up "later".

The same goes for splitting these up into separate tests and trying to make it a bit more readable. Maybe one of these days I'll find some time. 🙈  I don't find `spawn.mock.calls[0][2].detached` too clear. I'm thinking to aim towards something like:
```js
beforeEach(() => {
  processMock.platform = originalPlatform;
}

it('should spawn Jest in an attached process on Windows (not detached)', () => {
  processMock.platform = 'win32'

  const workspace: any = {pathToJest: ''};
  const args = [];
  createProcess(workspace, args);

  expect(spawn).toHaveBeenCalledTimes(1);
  const spawnArgs = spawn.mock.calls[0];
  const [_command, _args, options] = spawnArgs;
  const actual = !!options.detached;

  const expected = false;
  expect(actual).toBe(expected);
});

it('should spawn Jest in a detatched process on non-Windows systems', () => {
  processMock.platform = 'darwin'

  const workspace: any = {pathToJest: ''};
  const args = [];
  createProcess(workspace, args);

  expect(spawn).toHaveBeenCalledTimes(1);
  const spawnArgs = spawn.mock.calls[0];
  const [_command, _args, options] = spawnArgs;
  const actual = options.detached;

  const expected = true;
  expect(actual).toBe(expected);
});
```

src/__tests__/runner.test.js
```diff
@@ -323,15 +334,31 @@ describe('Runner', () => {
...
-    it('calls kill() to close the process on POSIX', () => {
+    it('calls process.kill() with processGroup id to close the process on POSIX', () => {
+      const workspace: any = {};
+      const sut = new Runner(workspace);
+      process.platform = 'posix';
+      const kill = jest.fn();
+      sut.debugprocess = ({kill, pid: 123}: any);
+      sut.closeProcess();
+
+      expect(kill).not.toBeCalled();
```
seanpoulter, line 345: Non-blocking: This is another test we should **eventually** clean up to communicate the behaviour we're expecting. If this failed, I wouldn't know what was wrong unless I looked at the function under test to see that `processKill` somehow threw an error. (Like the next test 😁).

seanpoulter (approved):
That was quick work @connectdotz! 👏

jeantil:
> I have only tested it on Windows, but to me the PR looks fine. Maybe someone should still test it on Linux, just to be sure.

I run on linux and I am impacted by the issue, I'm interested in testing the fix but I haven't the faintest idea how to do that ...

connectdotz:
@jeantil it's true that this is hard to test in isolation... I usually test it with `vscode-jest`, which we should cut a release ASAP then maybe you can help us to test the release candidate there? 

@seanpoulter sorry I didn't see your review until now, I will address them and cut a release later after work.

Also, can you guys review the other leg of this issue in jest-community/vscode-jest#573 ? Hopefully we can cut a `vscode-jest` release candidate this weekend...

jeantil:
I had never looked in details at how extensions are packaged but it looks like it's only two files produced by webpack. Building master with this PR included seems a bit involved but if you have a built extension.js I don't mind dropping it in place of the current one and restarting vscode.  I encounter the orphan process issue very often I will let you know really quick if this fixes it and in which cases :)

connectdotz:
@jeantil 

here is a .vsix file from my local build (had to zip it as github won't accept .vsix file in comments)

[vscode-jest-3.2.0.vsix.zip](https://github.com/jest-community/jest-editor-support/files/4531900/vscode-jest-3.2.0.vsix.zip)

you can follow vscode's instruction to install: https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix

Thanks for your help! If you don't feel comfortable taking a binary this way, I can fully understand... Hopefully, we should be able to produce the official candidate in a few days.

connectdotz:
just realized @seanpoulter volunteered to refactor the tests later, sounds good I will leave them to you then. merging...

jeantil:
I can confirm that using the provided vsix on linux the orphan processes issue disappeared

PR #42, opened 2020-05-01
Removed incorrect semicolon in type definition

xlarsx:


seanpoulter (approved):
Looks good. Thanks Luis. 👏 

How'd you find that? It seems like something that we should add to the build tooling and/or a Git hook.

xlarsx:
> How'd you find that? It seems like something that we should add to the build tooling and/or a Git hook.

When trying to compile a Typescript project that depends on this library it was throwing an error 😊


Thanks for the feedback

PR #45, opened 2020-05-05
Renamed pathToJest to jestCommandLine in ProjectWorkspace

### LINKED ISSUE #44: Improvement: Rename and document `pathToJest`
I'm creating this issue to keep track of the suggestion from #41 and set a v4.0 milestone.

**Acceptance criteria**

1. > https://github.com/jest-community/jest-editor-support/issues/41#issuecomment-621270042:
    > We could just rename the `ProjectWorkspace.pathToJest` to `ProjectWorkspace.jestCommandline`
2. Document the API change
3. (Suggested) Continue to support `pathToJest` as a deprecated value.

<br/>

**CC** @connectdotz @rossknudsen

rossknudsen:
Should close #44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Bug-fixes within the same version aren't needed
 * Replace babylon and typescript parsers with @babel/parser 7.x - @firsttris 
 -->
 
+### 28.0.0
+* Renamed the property `pathToJest` to `jestCommandLine` in the ProjectWorkspace configuration object to better convey how the property is used internally.  Left the original `pathToJest` with a deprecated flag.
+
 ### 27.2.0
 * Address orphan process issue - @connectdotz
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Bug-fixes within the same version aren't needed
 * Replace babylon and typescript parsers with @babel/parser 7.x - @firsttris 
 -->
 
-### 28.0.0
+### 27.3.0
 * Renamed the property `pathToJest` to `jestCommandLine` in the ProjectWorkspace configuration object to better convey how the property is used internally.  Left the original `pathToJest` with a deprecated flag.
 
 ### 27.2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,8 @@ Bug-fixes within the same version aren't needed
 
 ## Master
 * Replace babylon and typescript parsers with @babel/parser 7.x - @firsttris 
--->
-
-### 27.3.0
 * Renamed the property `pathToJest` to `jestCommandLine` in the ProjectWorkspace configuration object to better convey how the property is used internally.  Left the original `pathToJest` with a deprecated flag.
+-->
 
 ### 27.2.0
 * Address orphan process issue - @connectdotz
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,13 +39,14 @@ export function getSettings(workspace: ProjectWorkspace, options?: Options): Pro
 export class ProjectWorkspace {
   constructor(
     rootPath: string,
-    pathToJest: string,
+    jestCommandLine: string,
     pathToConfig: string,
-    localJestMajorVersin: number,
+    localJestMajorVersion: number,
     outputFileSuffix?: string,
     collectCoverage?: boolean,
     debug?: boolean,
   );
+  jestCommandLine: string;
   pathToJest: string;
   pathToConfig: string;
   rootPath: string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,6 +36,18 @@ export interface JestSettings {
 
 export function getSettings(workspace: ProjectWorkspace, options?: Options): Promise<JestSettings>;
 
+export function createProjectWorkspace(config: ProjectWorkspaceConfig): ProjectWorkspace;
+
+export interface ProjectWorkspaceConfig {
+  jestCommandLine: string;
+  pathToConfig?: string;
+  rootPath: string;
+  localJestMajorVersion: number;
+  outputFileSuffix?: string;
+  collectCoverage?: boolean;
+  debug?: boolean;
+}
+
 export class ProjectWorkspace {
   constructor(
     rootPath: string,
@@ -47,7 +59,6 @@ export class ProjectWorkspace {
     debug?: boolean,
   );
   jestCommandLine: string;
-  pathToJest: string;
   pathToConfig: string;
   rootPath: string;
   localJestMajorVersion: number;
--- a/src/Process.js
+++ b/src/Process.js
@@ -18,7 +18,7 @@ import ProjectWorkspace from './project_workspace';
  */
 // eslint-disable-next-line import/prefer-default-export
 export const createProcess = (workspace: ProjectWorkspace, args: Array<string>): ChildProcess => {
-  const runtimeExecutable = [workspace.jestCommandLine || workspace.pathToJest, ...args];
+  const runtimeExecutable = [workspace.jestCommandLine, ...args];
 
   // If a path to configuration file was defined, push it to runtimeArgs
   if (workspace.pathToConfig) {
--- a/src/Process.js
+++ b/src/Process.js
@@ -18,7 +18,7 @@ import ProjectWorkspace from './project_workspace';
  */
 // eslint-disable-next-line import/prefer-default-export
 export const createProcess = (workspace: ProjectWorkspace, args: Array<string>): ChildProcess => {
-  const runtimeExecutable = [workspace.pathToJest, ...args];
+  const runtimeExecutable = [workspace.jestCommandLine || workspace.pathToJest, ...args];
 
   // If a path to configuration file was defined, push it to runtimeArgs
   if (workspace.pathToConfig) {
--- a/src/__tests__/process.test.js
+++ b/src/__tests__/process.test.js
@@ -18,11 +18,20 @@ describe('createProcess', () => {
   });
 
   it('spawns the process', () => {
-    const workspace: any = {pathToJest: ''};
+    const workspace: any = {jestCommandLine: ''};
     const args = [];
     createProcess(workspace, args);
 
     expect(spawn).toBeCalled();
+  });
+
+  it('spawns the command from workspace.jestCommandLine', () => {
+    const workspace: any = {jestCommandLine: 'jest'};
+    const args = [];
+    createProcess(workspace, args);
+
+    expect(spawn.mock.calls[0][0]).toBe('jest');
+    expect(spawn.mock.calls[0][1]).toEqual([]);
   });
 
   it('spawns the command from workspace.pathToJest', () => {
@@ -34,8 +43,17 @@ describe('createProcess', () => {
     expect(spawn.mock.calls[0][1]).toEqual([]);
   });
 
-  it('spawns a command with spaces from workspace.pathToJest', () => {
-    const workspace: any = {pathToJest: 'npm test --'};
+  it('spawns the command from workspace.jestCommandLine ignoring workspace.pathToJest', () => {
+    const workspace: any = {jestCommandLine: 'jest', pathToJest: 'not jest'};
+    const args = [];
+    createProcess(workspace, args);
+
+    expect(spawn.mock.calls[0][0]).toBe('jest');
+    expect(spawn.mock.calls[0][1]).toEqual([]);
+  });
+
+  it('spawns a command with spaces from workspace.jestCommandLine', () => {
+    const workspace: any = {jestCommandLine: 'npm test --'};
     const args = [];
     createProcess(workspace, args);
 
@@ -43,7 +61,7 @@ describe('createProcess', () => {
   });
 
   it('appends args', () => {
-    const workspace: any = {pathToJest: 'npm test --'};
+    const workspace: any = {jestCommandLine: 'npm test --'};
     const args = ['--option', 'value', '--another'];
     createProcess(workspace, args);
 
@@ -53,7 +71,7 @@ describe('createProcess', () => {
   it('sets the --config arg to workspace.pathToConfig', () => {
     const workspace: any = {
       pathToConfig: 'non-standard.jest.js',
-      pathToJest: 'npm test --',
+      jestCommandLine: 'npm test --',
     };
     const args = ['--option', 'value'];
     createProcess(workspace, args);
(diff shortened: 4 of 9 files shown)


rossknudsen:
I've updated the changelog, bumping the major version since there a breaking change in the public API.

src/project_workspace.ts
```diff
@@ -69,15 +76,17 @@ export default class ProjectWorkspace {
 
   constructor(
     rootPath: string,
-    pathToJest: string,
-    pathToConfig: string,
+    jestCommandLine: string,
     localJestMajorVersion: number,
+    pathToConfig?: string,
```
connectdotz: if we agree on the shadow access approach, then we definitely do not want to change the order of these arguments...

src/project_workspace.ts
```diff
@@ -18,24 +18,31 @@ export default class ProjectWorkspace {
...
+  jestCommandLine: string;
 
   /**
-   * Path to a local Jest config file.
+   * @deprecated please use `jestCommandLine` instead.  If both settings are provided, only
+   * `jestCommandLine` will be used.
+   *
+   * @type {string?}
+   */
+  pathToJest?: string;
```
connectdotz: instead of maintaining both `jestCommandLine` and `pathToJest` internally and separately, what do you think if we just have shadow accessors (getter/setter) for `pathToJest` that backed by the one and only `jestCommandLine` property in the `ProjectWorkspace`. All codes/types can then just move to adopt the new name without checking both properties...

connectdotz (commented):
@rossknudsen sorry took a while to respond, thanks for making this change! 👍 

Before we fully migrate to ts, still need to maintain `index.d.ts` by hand, please make sure to update that also. 

Had some questions below, let's discuss...

src/project_workspace.ts
rossknudsen: Yeah that sounds good. Getters and setters didn't occur to me for use in JS. I think it will solve this problem, maintaining the constructor as is.  Will have a look at it.

src/project_workspace.ts
rossknudsen: I remember why I wanted to change the order of these arguments.  Its because the `pathToConfig` property is genuinely optional.  It is checked before it is used and it makes sense that there are occasions when there is no config file - just use the Jest defaults.  For the purposes of this PR, I have reverted the constructor to maintain backwards compatibility.  I will probably aim to tackle this scenario in the TS conversion PR.  I personally think that we should use an interface for this - which is easier to do in TS.

To work around this issue in my code I do this: `pathToConfig: undefined as unknown as string`.

src/project_workspace.ts
connectdotz: you can even consider adding some warning messages in the getter/setter to remind people to migrate...

rossknudsen: Do you mean `console.warn(...)`?

src/project_workspace.ts
seanpoulter: Just a thought -- If we're having to worry about argument order, that signals to me we should consider using an Object.

rossknudsen: Yes, I agree that an object would be the best option here.  But my opinion is that it will be easier to make that transition once more of the code is converted to Typescript.  I just played with defining an interface for the ProjectWorkspace class, and then realized that I would need to update some existing code in JS with Flow typings...

src/project_workspace.ts
connectdotz: yes

src/Process.js
```diff
@@ -18,7 +18,7 @@ import ProjectWorkspace from './project_workspace';
  */
 // eslint-disable-next-line import/prefer-default-export
 export const createProcess = (workspace: ProjectWorkspace, args: Array<string>): ChildProcess => {
-  const runtimeExecutable = [workspace.pathToJest, ...args];
+  const runtimeExecutable = [workspace.jestCommandLine || workspace.pathToJest, ...args];
```
connectdotz: do we still need to check `pathToJest` here?

src/__tests__/project_workspace.test.js
```diff
@@ -11,9 +11,9 @@ import ProjectWorkspace from '../project_workspace';
 
 describe('setup', () => {
```
connectdotz, line 12: should we add a test to ensure that `pathTojest` is just a shadow accessor of `jestCommandLine` and not its own property?

src/Process.js
connectdotz: do we still need to check `pathToJest` here? The whole idea of using shadow accessor is so that the rest of the system doesn't need to concern itself with `pathToJest` anymore, as far as they concern, everything has changed to `jestCommandLine`...

src/__tests__/process.test.js
```diff
@@ -34,16 +43,25 @@ describe('createProcess', () => {
     expect(spawn.mock.calls[0][1]).toEqual([]);
   });
 
-  it('spawns a command with spaces from workspace.pathToJest', () => {
-    const workspace: any = {pathToJest: 'npm test --'};
+  it('spawns the command from workspace.jestCommandLine ignoring workspace.pathToJest', () => {
+    const workspace: any = {jestCommandLine: 'jest', pathToJest: 'not jest'};
```
connectdotz: this is not a valid workspace, we should not have `pathToJest` property in workspace any more...

src/__tests__/project_workspace.test.js
connectdotz, line 12: this is the place I think we should test `pathToJest` is a shadow accessor instead of a real concrete property, while all the other classes should only deal with `jestCommandLine`.

index.d.ts
```diff
@@ -39,13 +39,14 @@ export function getSettings(workspace: ProjectWorkspace, options?: Options): Pro
...
+    jestCommandLine: string,
     pathToConfig: string,
-    localJestMajorVersin: number,
+    localJestMajorVersion: number,
     outputFileSuffix?: string,
     collectCoverage?: boolean,
     debug?: boolean,
   );
+  jestCommandLine: string;
   pathToJest: string;
```
connectdotz: given this is a type file that it won't impact runtime (the binary out there that uses pathToJest will continue to work), we could get rid of `pathToJest`, so our customers will see ts warning during development and will need to fix the signature in their next release.

connectdotz (commented):
thanks for the update, a few comments below.

I think basically we are trying to push for runtime backward compatibility (for our client) while transitioning over to the new name. In order to minimize the complexity of checking both properties within this package, we adopted the shadow accessors so the rest of the system can safely focus only on the new property, i.e. none of the classes/functions outside of `ProjectWorkspace` should consider `pathToJest`...

src/project_workspace.ts
connectdotz: actually come to think about this, why is this a class? what does it do other than holding some config data? maybe in the next major version, possible typescript conversion (?), we can rewrite this...

src/__tests__/project_workspace.test.js
seanpoulter: That sounds like it might be fragile if we end up testing implementation details. Can we do that without knowing the internals?

Given a new ProjectWorkspace created with a jestCommandLine:
  - it should return the jestCommandLine
    ```js
    const commandLine = {} /* some arg */
    const instance = new ProjectWorkspace(..., commandLine, ...);
    const expected = commandLine;
    const actual = instance.jestCommandLine;

    expect(actual).toBe(expected);
    ```
  - it should continue to support the deprecated pathToJest
    ```js
    // The same setup with
    const actual = instance.jestCommandLine;

    expect(actual).toBe(expected);
    ```

Given a new ProjectWorkspace created with a pathToJest:
  - it should return the jestCommandLine
  - it should continue to support the deprecated pathToJest

src/project_workspace.ts
seanpoulter: Sorry to derail that thread. I've created #46 to capture that idea. We can defer that for now.

To get back to the original point, the API is messy. I'd say leave the argument order for now and defer changing the API signature. In the interim Ross, you could build a factory function that only has to add that `pathToConfig: undefined as unknown as string` mess once.

```js
const workspace = ({ ... }) => new ProjectWorkspace( ... );
```

src/Process.js
rossknudsen: I was kind of forward thinking here, that if we were to replace the workspace config with an object, then the pathToJest might have been supplied.  But it does come back to timing, maybe the pathToJest will be removed then.  Happy to remove it.

src/__tests__/project_workspace.test.js
rossknudsen: I had a test which exercised this functionality (at least IMO).  I moved it from the settings tests to this file.  Here is the definition:

```ts
it('ensure that pathToJest is a shadow accessor for jestCommandLine', () => {
    const localJestMajorVersion = 1000;
    const pathToConfig = 'test';
    const jestCommandLine = 'path_to_jest';
    const rootPath = 'root_path';
    const workspace = new ProjectWorkspace(rootPath, jestCommandLine, pathToConfig, localJestMajorVersion);

    // first check that both pathToConfig and jestCommandLine are the same value.
    expect(workspace.jestCommandLine).toBe(jestCommandLine);
    expect(workspace.pathToJest).toBe(jestCommandLine);

    // then update the pathToJest property.
    const updatedPathToJest = 'updated pathToJest';
    workspace.pathToJest = updatedPathToJest;

    // check that both pathToConfig and jestCommandLine yield the same value.
    expect(workspace.jestCommandLine).toBe(updatedPathToJest);
    expect(workspace.pathToJest).toBe(updatedPathToJest);
  });
```
Note that it is not possible to 'create' a config object with pathToJest rather than jestCommandLine because the constructor does not allow it (we should test this if/when we have a config _object_).  To test the shadow access, I have used property assignment instead.  Its unlikely anyone would create the object and then reassign the properties but it does prove the desired implementation.

src/Process.js
rossknudsen: Actually in the light of the rest of this conversation, I'll remove this.  If/when we move to a config object, then we might need to check both fields - depending on how long we want to maintain backwards compatibility.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
       /* Basic Options */
       "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
       "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
-      "lib": ["ES2015"],                        /* Specify library files to be included in the compilation. */
+      "lib": ["DOM", "ES2015"],                 /* Specify library files to be included in the compilation. */
```
connectdotz: noticed you added 'DOM' here, is this intentional?

src/project_workspace.ts
```diff
@@ -6,6 +6,16 @@
  *
  */
 
+export interface ProjectWorkspaceConfig {
+  jestCommandLine: string;
+  pathToConfig?: string;
```
connectdotz, line 11: I hate to mention this now, but if we are going for jestCommandLine, do you think we should also retire `pathToConfig`? Sorry, I should have thought about this earlier...

connectdotz (commented):
thanks for the update, see comments below

tsconfig.json
rossknudsen: The `console` global is defined in the DOM lib.  I'm using the console to emit warnings when using the deprecated properties.

src/project_workspace.ts
rossknudsen: IMO I think the `pathToConfig` property is still needed. It just needs to be optional rather than required. Since I was introducing the `ProjectWorkspaceConfig` interface, which hopefully will be the future, I thought I'd mark `pathToConfig` as optional from the get go.

tsconfig.json
connectdotz: I see... based on tsconfig [doc](https://www.typescriptlang.org/docs/handbook/compiler-options.html):

> Note: If --lib is not specified a default list of libraries are injected. The default libraries injected are:
► For --target ES5: DOM,ES5,ScriptHost
► For --target ES6: DOM,ES6,DOM.Iterable,ScriptHost

seems like target `es6` by default already include DOM, in addition to `es2015`, maybe we don't need to specify `lib` at all?

Anyway, it's not wrong just a bit redundant...

rossknudsen: I'm fine with that.  Just commented the line to match the rest of the config file.  We'll consolidate that config once everything is converted to TS.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Bug-fixes within the same version aren't needed
 * Replace babylon and typescript parsers with @babel/parser 7.x - @firsttris 
 -->
 
+### 27.3.0
```
connectdotz: actually no need to specify the version number, just add the change in the commented out `master` section above.  When we are ready to cut the release, they will be moved to the proper version then...

tsconfig.json
```diff
@@ -3,7 +3,7 @@
       /* Basic Options */
       "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
       "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
-      "lib": ["ES2015"],                        /* Specify library files to be included in the compilation. */
+      // "lib": ["DOM", "ES2015"],              /* Specify library files to be included in the compilation. */
       "allowJs": false,                         /* Allow javascript files to be compiled. */
       // "checkJs": true,                       /* Report errors in .js files. */
       // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
```
connectdotz: please remove the commented out statements

src/project_workspace.ts
connectdotz: I am not sure what the use case will be... but I am fine in leaving it in for now.

connectdotz (approved):
just a few minor clean up, we are ready to merge once done.

CHANGELOG.md
rossknudsen: Ok, cool.

tsconfig.json
rossknudsen: Done

connectdotz:
@rossknudsen thanks for the PR 👍 , I will cut a beta soon so we can start to play with the new changes

rossknudsen:
Sounds good, I was starting to see some parsing issues around it.each, so I'm keen to see how the new one works. I'll update my other TS PR too - whether it makes this release or the next one.

PR #52, opened 2020-11-18
fix return statement without argument crash

connectdotz:
fix parsing error for  no argument return statement: `return;`

### DIFF
--- a/src/parsers/babel_parser.js
+++ b/src/parsers/babel_parser.js
@@ -161,7 +161,7 @@ export const parse = (file: string, data: ?string, options: ?parser.ParserOption
         isFunctionDeclaration(element.expression.right.type)
       ) {
         searchNodes(element.expression.right.body, parent);
-      } else if (element.type === 'ReturnStatement' && element.argument && element.argument.arguments) {
+      } else if (element.type === 'ReturnStatement' && element.argument?.arguments) {
         element.argument.arguments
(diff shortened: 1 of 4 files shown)


connectdotz:
@firsttris do you have a few minutes to review this?

src/parsers/babel_parser.js
```diff
@@ -161,7 +161,7 @@ export const parse = (file: string, data: ?string, options: ?parser.ParserOption
         isFunctionDeclaration(element.expression.right.type)
       ) {
         searchNodes(element.expression.right.body, parent);
-      } else if (element.type === 'ReturnStatement' && element.argument.arguments) {
+      } else if (element.type === 'ReturnStatement' && element.argument && element.argument.arguments) {
```
firsttris: `else if (element.type === 'ReturnStatement' && element.argument?.arguments)`

you could use optional chaining here instead.
even if its still javascript it's already used a few times in this file.

firsttris (commented):
of course it's a pleasure

src/parsers/babel_parser.js
connectdotz: good point

connectdotz:
regarding coverage decrease: looking at the detailed report, the `src` coverage has not changed but the overall coverage reduced due to the removal of `fixures`, which should not have been included anyway.

PR #58, opened 2020-12-29
Bugfix: Recognize `it.only.each` and other chains off of it.

### LINKED ISSUE #53: Parsing fails when using test.each`table`(name, fn, timeout)
From the Jest docs, you can structure a describe or a test like so
```typescript
test.each`
  a    | b    | expected
  ${1} | ${1} | ${2}
  ${1} | ${2} | ${3}
  ${2} | ${1} | ${3}
`('returns $expected when $a is added $b', ({a, b, expected}) => {
  expect(a + b).toBe(expected);
});
```
[reference](https://jestjs.io/docs/en/api#2--testeachtablename-fn-timeout-)

jest-editor-support does not recognize these expressions as tests in either v28.0.0 or v27.2.0.
I think this is similar, but different from the discussion had in https://github.com/jest-community/jest-editor-support/issues/29#issuecomment-578521119

### LINKED ISSUE #59: Does not parse `it` when used with more than one modifier
The Jest API includes a handful of modifiers to `it` that can be combined: `only`/`skip`, `concurrent`, and `each`.  Currently, the parser will detect an `it` if it has, at most, one modifier on it.  This means statements like `it.concurrent.each` are not detected.

Unsupported Jest API methods (from https://jestjs.io/docs/en/api):
* `describe.only.each(table)(name, fn)`
* `describe.skip.each(table)(name, fn)`
* ``describe.each`table`(name, fn)`` (and all variations above)
* `test.concurrent.each(table)(name, fn, timeout)`
* `test.concurrent.only.each(table)(name, fn)`
* `test.concurrent.skip.each(table)(name, fn)`
* `test.only.each(table)(name, fn)`
* `test.skip.each(table)(name, fn)`
* ``test.each`table`(name, fn)`` (and all variations above)

TheSench:
This PR enhances parsing of `it.each` and `describe.each` by allowing them to also be found when deeper in a property chain (e.g. `it.only.each`, `it.concurrent.only.each`, `describe.only.each`).  In addition, it also allows the **Tagged Template Literal** syntax of `it.each` and `describe.each` to be parsed.

It updates the logic that identifies `it` and `describe` blocks to walk up the callee chain to the root when there are multiple levels of callee (e.g. for `it.each(scenarios)(name, fn)`), and then to walk up the tag and object tree to the root (e.g. `each`->`only`->`concurrent`->`it`).

Jest API methods that should now be supported (from https://jestjs.io/docs/en/api):
* `describe.only.each(table)(name, fn)`
* `describe.skip.each(table)(name, fn)`
* ``describe.each`table`(name, fn)`` (and all variations above)
* `test.concurrent.each(table)(name, fn, timeout)`
* `test.concurrent.only.each(table)(name, fn)`
* `test.concurrent.skip.each(table)(name, fn)`
* `test.only.each(table)(name, fn)`
* `test.skip.each(table)(name, fn)`
* ``test.each`table`(name, fn)`` (and all variations above)

Fixes #53 
Fixes #59

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
-* Detect Tagged Template Literal version of describe.each and it.each
-* Detect it when used with deep chain of modifiers ( e.g. `test.concurrent.only.each(table)(name, fn)`)
+* Detect Tagged Template Literal version of describe.each and it.each - @TheSench
+* Detect it when used with deep chain of modifiers (e.g. `test.concurrent.only.each(table)(name, fn)`) - @TheSench
 -->
 
 ### 28.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
+* Detect Tagged Template Literal version of describe.each and it.each
+* Detect it when used with deep chain of modifiers ( e.g. `test.concurrent.only.each(table)(name, fn)`)
 -->
 
 ### 28.0.0
--- a/fixtures/parser_tests.js
+++ b/fixtures/parser_tests.js
@@ -25,9 +25,10 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
       }).not.toThrow();
     });
   });
-  /**
-   * Simplify parseResult to make diffing easier during failed tests
-   */
+  const parseResultNames = (parseResult: ParseResult) => ({
+    itBlocks: parseResult.itBlocks.map(it => it.name),
+    describeBlocks: parseResult.describeBlocks.map(describe => describe.name)
+  });
   const simplifiedParseResult = (parseResult: ParseResult) => ({
     itBlocks: parseResult.itBlocks.map(it => ({
       name: it.name,
@@ -39,7 +40,18 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
       start: describe.start,
       end: describe.end
     }))
-  })
+  });
+  /**
+   * Compare names, start, and end of each it/describe in parse result
+   */
+  const assertParseResultSimple = (parseResult: ParseResult, expectedSimplifiedParseResult) => {
+    // Check just names first
+    const names = parseResultNames(parseResult);
+    const expectedNames = parseResultNames(expectedSimplifiedParseResult);
+    expect(names).toMatchObject(expectedNames);
+    // Then check names, start, and end
+    expect(simplifiedParseResult(parseResult)).toMatchObject(expectedSimplifiedParseResult);
+  };
 
   describe('File Parsing for it blocks', () => {
     it('For the simplest it cases', () => {
@@ -404,8 +416,7 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
     it('For the simplest it.each cases', () => {
       const parseResult = parse(`${fixtures}/global_it_eaches.example`);
 
-      const filteredParseResult = simplifiedParseResult(parseResult);
-      expect(filteredParseResult).toMatchObject({
+      assertParseResultSimple(parseResult, {
         itBlocks: [
           {
             name: 'works with old functions',
@@ -479,16 +490,19 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
           }
         ],
         describeBlocks: [
+          // No describes
         ]
       });
     });
 
     it('For the simplest describe.each cases', () => {
-      const parseResult = parse(`${fixtures}/global_describe_eaches.example`);
+      const parseResult = parse(`${fixtures}/describe_eaches.example`);
 
-      const filteredParseResult = simplifiedParseResult(parseResult);
-      expect(filteredParseResult).toMatchObject({
+      assertParseResultSimple(parseResult, {
         itBlocks: [
+          // No tests
+        ],
+        describeBlocks: [
           {
             name: 'works with old functions',
             start: { line: 2, column: 1 },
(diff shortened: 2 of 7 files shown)


src/parsers/babel_parser.js
```diff
@@ -30,6 +30,20 @@ export const parse = (file: string, data: ?string, options: ?parser.ParserOption
...
+      rootForType = rootForType[type];
+    }
+    return rootForType;
+  };
+
+  const getRootCalleeOrSelf = node => getRootOfTypeOrSelf(node, 'callee');
+
+  const getRootTagOrSelf = node => getRootOfTypeOrSelf(node, 'tag');
+
+  const getRootObjectOrSelf = node => getRootOfTypeOrSelf(node, 'object');
```
connectdotz: I like your approach but it seems that we could make it more expressive...

```suggestion
  const deepGet = (node, ...types: string[]) => 
    types.reduce<BabelNode>((rootForType, type) => {
      while (rootForType[type]) {
        rootForType = rootForType[type];
      }
      return rootForType;
    }, node);
```

so we can call this with clear order below...

src/parsers/babel_parser.js
```diff
@@ -76,11 +90,9 @@ export const parse = (file: string, data: ?string, options: ?parser.ParserOption
...
   // handle cases where it's a member expression (.only)
   const getNameForNode = node => {
-    if (isFunctionCall(node) && node && node.expression && node.expression.callee) {
-      const name =
-        node.expression.callee.name ||
-        node.expression.callee.object?.name ||
-        node.expression.callee.callee?.object?.name;
+    if (isFunctionCall(node) && node.expression.callee) {
+      const rootCallee = getRootCalleeOrSelf(node.expression);
+      const name = rootCallee.name || getRootObjectOrSelf(getRootTagOrSelf(rootCallee)).name;
```
connectdotz: use the `deepGet` above, we can express the order and attributes in a more readable way... (I know, I know, "readability" is rather subjective...)

```suggestion
      const callee = deepGet(node.expression, 'callee');
      const name =
        callee.name ||
        deepGet(callee, 'object').name ||
        deepGet(callee, 'tag').name ||
        deepGet(callee, 'tag', 'object').name;
```

connectdotz (commented):
hey @TheSench thank you for submitting this PR, this is definitely a much-needed fix 👍 

I agree with the overall approach and your code does work. However, I think we can make the code more "expressive" for showing precedence and properties used. I made some suggestions below please take a look and see if it makes sense.

src/parsers/babel_parser.js
TheSench: > I like your approach but it seems that we could make it more expressive...
> 
> so we can call this with clear order below...

I like this suggestion.

src/parsers/babel_parser.js
TheSench: > use the `deepGet` above, we can express the order and attributes in a more readable way... (I know, I know, "readability" is rather subjective...)

I've added this, except I did not include `deepGet(callee, 'tag').name`, as that's not a valid case (`describe` or `it` themselves are never a tag).

connectdotz: good point

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
+* Detect Tagged Template Literal version of describe.each and it.each
+* Detect it when used with deep chain of modifiers ( e.g. `test.concurrent.only.each(table)(name, fn)`)
```
connectdotz: as a convention, please attach the author's name at the end of the changelog.

fixtures/parser_tests.js
```diff
@@ -385,6 +385,52 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
       assertBlock2(itBlock, 2, 7, 4, 9, 'each test %p');
       assertNameInfo(itBlock, 'each test %p', 2, 33, 2, 44);
     });
+    
+    it('For the simplest it.each cases', () => {
```
connectdotz: Seems like we are missing the corresponding `describe.each` test cases? Also might be worthy to add nested ".each" cases ...

fixtures/parser_tests.js
```diff
@@ -401,6 +447,75 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
       assertBlock2(itBlock, 2, 7, 6, 9, 'each test %p');
       assertNameInfo(itBlock, 'each test %p', 3, 10, 3, 21);
     });
+
+    it('For tagged template syntax', () => {
```
connectdotz, line 701: good this one has both test and describe. Same comment as above, let's add some nested `.each` test cases to be thorough

src/parsers/babel_parser.js
```diff
@@ -74,13 +83,16 @@ export const parse = (file: string, data: ?string, options: ?parser.ParserOption
...
+      const rootCallee = deepGet(node.expression, 'callee');
       const name =
-        node.expression.callee.name ||
-        node.expression.callee.object?.name ||
-        node.expression.callee.callee?.object?.name;
+        rootCallee.name ||
+        // handle cases where it's a member expression (e.g .only or .concurrent.only)
+        deepGet(rootCallee, 'object').name ||
+        // handle cases where it's part of a tag (e.g. .each`table`)
+        deepGet(rootCallee, 'tag', 'object').name;
```
connectdotz, line 95: great to add these comments! 👍

connectdotz (commented):
Awesome, we are really close now. Just a few minor points about tests and changelog...

fixtures/parser_tests.js
TheSench: It added a similar describe.each suite.

fixtures/parser_tests.js
TheSench: I added a separate test fixture for some nested cases.

fixtures/parser_tests.js
```diff
@@ -385,51 +412,274 @@ function parserTests(parse: (file: string) => ParseResult, isTypescript = false)
...
 
-      const fifthIt = data.itBlocks[4];
-      expect(fifthIt.name).toEqual('works with it.only');
-      expect(fifthIt.start).toEqual({column: 1, line: 18});
-      expect(fifthIt.end).toEqual({column: 3, line: 20});
+      assertParseResultSimple(parseResult, {
+        itBlocks: [
+          {
+            name: 'works with old functions',
+            start: { line: 2, column: 1 },
```
connectdotz, line 423: looks like there is some prettier format inconsistency in this file... such as extra space in {}... 

Looks like ci didn't lint any file in fixtures thus missed it...

connectdotz (approved):
@TheSench thanks for the update, I did notice some prettier formate related issues. No big deal, I can fix it during merging.

Once again, thank you for the PR 👍

PR #63, opened 2021-03-01
fix snapshot, runner, test_reconciler

connectdotz:
change summary
## typescript types
- fix typescript types for coverage  

## Runner
  - adding extra arguments for more granular control of the run process
  - allow passing node_env 
  - emit close and exit events separately
  
## test_reconciler
  - add API to remove file from cache
  
## snapshot
- fixes snapshot parsing for flow and typescript files
- refactor and clean up parser modules and tests
- upgrade "jest-snapshot" dependency

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,6 +10,10 @@ import {ChildProcess} from 'child_process';
 import {Config as JestConfig} from '@jest/types';
 import { CoverageMap, CoverageMapData } from 'istanbul-lib-coverage';
 
+export interface RunArgs {
+  args: string[];
+  replace?: boolean; // default is false
+}
 export interface Options {
   createProcess?: (
     workspace: ProjectWorkspace,
@@ -19,10 +23,8 @@ export interface Options {
   testNamePattern?: string;
   testFileNamePattern?: string;
   reporters?: string[];
-  // extra args if specified will be appended to auto arguments
-  extraArgs?: string[];
-  // custom args if specified will supercede any auto arguement logic
-  args?: string[];
+  /** either to append or replace the Runner process arguments */
+  args?: RunArgs;
 }
 
 export type RunnerEvent = 'processClose' | 'processExit' | 'executableJSON' | 'executableStdErr' | 'executableOutput' | 'terminalError';
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,6 +8,7 @@
 import {EventEmitter} from 'events';
 import {ChildProcess} from 'child_process';
 import {Config as JestConfig} from '@jest/types';
+import { CoverageMap, CoverageMapData } from 'istanbul-lib-coverage';
 
 export interface Options {
   createProcess?: (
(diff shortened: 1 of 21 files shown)


index.d.ts
```diff
@@ -18,6 +19,8 @@ export interface Options {
   testNamePattern?: string;
   testFileNamePattern?: string;
   reporters?: string[];
+  // extra args if specified will be appended to auto arguments
```
stephtr: Wouldn't it make sense to use JSDoc's `/** */`instead?

index.d.ts
```diff
@@ -18,6 +19,8 @@ export interface Options {
   testNamePattern?: string;
   testFileNamePattern?: string;
   reporters?: string[];
+  // extra args if specified will be appended to auto arguments
+  extraArgs?: string[];
```
stephtr: I would call it something like `autoArgs`, since they don't get used if `args` is set.

stephtr:
I'm really sorry, but currently I have hardly any spare time available for this project. Unfortunately, I also don't know when I'll have time to comment on the other PRs.

So far this PR looks good to me.

stephtr (approved):

index.d.ts
connectdotz: good point, I guess we could consolidate these 2 into 1, like:
```
interface RunArgs {
    args: string[];
    replace?: boolean; 
}
```

connectdotz:
@stephtr no need to be sorry, we all do what we can, I appreciate you are taking the time to review this. Personally, I found review adds tremendous value by simply providing a different perspective, but I also understand that it is a mental overhead for the reviewers. A review is indeed a gift that I think should never be taken for granted. Thanks!

index.d.ts
```diff
@@ -18,8 +23,8 @@ export interface Options {
   testNamePattern?: string;
   testFileNamePattern?: string;
   reporters?: string[];
-  // custom args if specified will supercede any auto arguement logic
-  args?: string[];
+  /** either to append or replace the Runner process arguments */
+  args?: RunArgs;
```
stephtr, line 27: Do we care about introducing a breaking change? If yes, we could also use `args?: RunArgs | string[];`.

One thing I noticed with `RunArgs` is that it will read as `options.args.args`. Maybe it would be better to have the `replace` property as part of `Options`? We could also call it something like `SkipDefaultArgs`.

stephtr (commented):
Thank you for your nice words! It is even more precious if someone grabs those review notes and incorporates them in new and even better ideas!

index.d.ts
connectdotz: yeah, I thought about `args?: RunArgs | string[]` but we still have the `options.args.args` situation while trying to get flow to work with this kind of type... Regarding moving the "SkipDefaultArgs" flag to the options itself, it won't achieve the goal that this flag should always accompany the args that replace it...therefore, my gut feeling is it might not be worthy to do a breaking change right now, as none of the alternatives is perfect either... maybe revisit this when we migrate the rest of the package to typescript.

PR #71, opened 2021-05-25
parser: add additional attributes and preserve describe/it block names

connectdotz:
This PR updated parser to expose more and accurate attributes for the NamedBlock to help consumers matching the jest test results better:

1. preserve the "name" of the NamedBlock. It was using markers before, which might not be helpful when we want to show users what block it is. (this might be a breaking change if people are using the markers for something...)
2. added a new optional attribute `lastProperty` to help identify if the block is a `each` block for example. The "last property" is the "each" in `describe.skip.each()` or "only" in `it.only()`.
3. added a new optional attribute `nameType` to help identify the block's "name" type, such as 'Literal', 'TemplateLiteral', 'CallExpression'. This not only helps identify if the name is a static literal string that should match the test results' name, it also helps people that used to rely on the type-marker used in the "name" field before. 
4. upgraded dependency `@babel/parser` to the latest 7.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ Bug-fixes within the same version aren't needed
 ## Master
 * convert `parser-test.js` to a true jest test: `babel-parser.test.ts` and convert to typescript. - @connectdotz
 * upgrade prettier and @babel/preset-typescript, fix lint errors. - @connectdotz
-* adding lastProperty and hasDynamicName attributes to NamedNode - @connectdotz
-* preserve test/describe block names regardless of its types, i.e. string or template-literal or function etc - @connectdotz
+* adding lastProperty and nameType attributes to NamedNode - @connectdotz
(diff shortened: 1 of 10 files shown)


connectdotz:
@firsttris I think you are using this package as well. Would like to hear your opinion on the potential breaking change for the "name" of the describe/it blocks... do you have any concern?

connectdotz:
ok, I made the change to record the name "type" so the information is not lost after this PR. It should be functional-compatible with the previous version, that removed one of my main concern for this change. 

@stephtr or @firsttris do you have some time to review this? There are other fixes waiting for these changes, appreciate it if we can move forward on this sooner than later...

connectdotz:
Looks like you guys are busy, that's all right. Feel free to let me know if you object to this change, otherwise, I will merge this in and cut a major release to be safe and give people time to upgrade.

firsttris:
all good from my side!

Issue #68, opened 2021-05-13
Fail to parse typescript type assertion syntax


gregoryzh:
Hi, i think  the parser fails at angular bracket type assertion syntax, is this an known issue?
![image](https://user-images.githubusercontent.com/78626456/118168262-84f8b180-b3dc-11eb-82d3-aaddf9549441.png)

ehaynes99:
The error thrown is `SyntaxError: Unterminated JSX contents` during `@babel/parser` here:

https://github.com/jest-community/jest-editor-support/blob/4b34a1850adb8f880aefb5cb31377ac32e3025da/src/parsers/babel_parser.js#L25

connectdotz:
yes, this is a known issue, fix is not hard: we are enabling "jsx" babel plugin for all file types, what we should do is only enable "jsx" for react file types.

It's a relatively simple change if anybody is interested. You can start from [here](https://github.com/jest-community/jest-editor-support/blob/4b34a1850adb8f880aefb5cb31377ac32e3025da/src/parsers/helper.ts#L16)

gregoryzh:
> yes, this is a known issue, fix is not hard: we are enabling "jsx" babel plugin for all file types, what we should do is only enable "jsx" for react file types.
> 
> It's a relatively simple change if anybody is interested. You can start from [here](https://github.com/jest-community/jest-editor-support/blob/4b34a1850adb8f880aefb5cb31377ac32e3025da/src/parsers/helper.ts#L16)

hi, created PR
#72

ehaynes99:
JSX is not specific to React; it's used by many different libraries.

connectdotz:
@ehaynes99 the current thinking is to only remove 'jsx' babel plugin from `.ts` files. All the other file extensions will continue to be able to use 'jsx' syntax/plugin. Does that address your concern? If not, please elaborate.

Based on these discussions (Microsoft/TypeScript#26489 and Microsoft/TypeScript#30503), typescript standard is to allow 'jsx' syntax only for `.tsx` files. Even though many users have asked for 'jsx' in `.ts` files, the request did not seem to be approved so far. So I think removing 'jsx' plugin from `.ts` file would be all right as it is aligned with the typescript standard.

ehaynes99:
I think it's safe to consider bracket-style type assertions and `.tsx` files to be mutually exclusive, as typescript itself was unable to solve for both in the same file, leading to the introduction of the `as` keyword.
[...]

PR #72, opened 2021-05-26
fix parser crashes on typescript angle-bracket.

gregoryzh:
fix issue for #68 
added new test case, 
separate some jsx test cases into its on it block

### DIFF
--- a/src/parsers/helper.ts
+++ b/src/parsers/helper.ts
@@ -23,18 +23,6 @@ export const supportedFileType = (filePath: string): 'ts' | 'js' | undefined => 
   return undefined;
 };
 
-/**
- * determine if the file is a react file(jsx, tsx);
- * @param filepath
- * @returns true or false
- */
-export const isReactFileType = (filePath: string): boolean => {
-  if (filePath.match(/\.[jt]sx$/)) {
-    return true;
-  }
-  return false;
-};
-
 export const plugins: ParserPlugin[] = [
   'asyncGenerators',
   'bigInt',
@@ -66,8 +54,10 @@ export const parseOptions = (filePath: string, strictMode = false): ParserOption
   const fileType = supportedFileType(filePath);
 
   let parserPlugins = plugins;
-  // only add jsx plugin if file is react type.
-  if (fileType && isReactFileType(filePath)) {
+
+  // Add jsx plugin for jsx, js and tsx file, but exclude it for ts file since jsx syntax is not surppoted in ts file.
+  // https://github.com/microsoft/TypeScript/issues/26489
+  if (fileType && !filePath.match(/\.ts$/i)) {
     parserPlugins = [...plugins, 'jsx'];
   }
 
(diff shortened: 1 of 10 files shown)


connectdotz (changes requested):
@gregoryzh thanks for taking on this issue 👍 

I think we only need to exclude "jsx" plugin for ".ts"  files, as I have seen ".js" file with react syntax so should continue to include "jsx" for them.

also please update CHANGELOG.md

thanks!

src/parsers/helper.ts
```diff
@@ -23,6 +23,18 @@ export const supportedFileType = (filePath: string): 'ts' | 'js' | undefined =>
...
+/**
+ * determine if the file is a react file(jsx, tsx);
+ * @param filepath
+ * @returns true or false
+ */
+export const isReactFileType = (filePath: string): boolean => {
+  if (filePath.match(/\.[jt]sx$/)) {
+    return true;
+  }
+  return false;
```
ehaynes99: The conditional return is the same as returning the condition. Also, you likely want case insensitivity with `/i`.
```suggestion
  return filePath.match(/\.[jt]sx$/i);
```

src/parsers/helper.ts
```diff
@@ -23,6 +23,18 @@ export const supportedFileType = (filePath: string): 'ts' | 'js' | undefined =>
   return undefined;
 };
 
+/**
+ * determine if the file is a react file(jsx, tsx);
+ * @param filepath
+ * @returns true or false
+ */
+export const isReactFileType = (filePath: string): boolean => {
```
ehaynes99: This is a misnomer. JSX is used for many other things than react files.

src/parsers/helper.ts
```diff
@@ -53,10 +64,17 @@ export const plugins: ParserPlugin[] = [
 
 export const parseOptions = (filePath: string, strictMode = false): ParserOptions | null => {
   const fileType = supportedFileType(filePath);
+
+  let parserPlugins = plugins;
+  // only add jsx plugin if file is react type.
+  if (fileType && isReactFileType(filePath)) {
```
ehaynes99: While the `x` appears to be enforced in Typescript, the overwhelming majority of javascript React files are simply `.js`, not `.jsx`.

`create-react-app` recommended against the `.jsx` extension in 2016. https://github.com/facebook/create-react-app/releases/tag/v0.4.1

src/parsers/helper.ts
gregoryzh: updated logic to check only against ts.

src/parsers/helper.ts
gregoryzh: removed the helper now, since it is not necessary when we now only checking against ts file.

src/parsers/helper.ts
gregoryzh: updated logic to only exclude jsx plugin for ts files.

gregoryzh:
> @gregoryzh thanks for taking on this issue 👍
> 
> I think we only need to exclude "jsx" plugin for ".ts" files, as I have seen ".js" file with react syntax so should continue to include "jsx" for them.
> 
> also please update CHANGELOG.md
> 
> thanks!

hi, updated as per commented, please take a look when you have chance

PR #75, opened 2021-06-04
Disable jsx plugin for .ts files

ehaynes99:
The JSX plugin for Typescript is not compatible with bracket-style type assertions. This led to the introduction of the `as` keyword as an alternate:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-6.html#new-tsx-file-extension-and-as-operator

This restricts the plugin to only be used with `.tsx` files.

NOTE: There was a copy/paste error in the babel parser test that was skipping over the existing JSX test. Since other tests in the same file were using a comparison of the whole object that precludes such omissions, I changed the style of 2 of the tests to be consistent with this.

### DIFF
--- a/src/parsers/helper.ts
+++ b/src/parsers/helper.ts
@@ -9,13 +9,17 @@
 import {ParserOptions, ParserPlugin} from '@babel/parser';
 
 /**
- * determine if the file is a typescript (ts), javascript (js) otherwise returns undefined.
+ * determine if the file is a javascript (js), typescript (ts), or typescript JSX (tsx),
+ * otherwise returns undefined.
  * @param filepath
- * @returns 'js'|'ts' or undefined
+ * @returns 'js'|'ts'|'tsx' or undefined
  */
-export const supportedFileType = (filePath: string): 'ts' | 'js' | undefined => {
-  if (filePath.match(/\.tsx?$/)) {
+export const getFileType = (filePath: string): 'js' | 'ts' | 'tsx' | undefined => {
+  if (filePath.match(/\.ts$/)) {
     return 'ts';
+  }
+  if (filePath.match(/\.tsx$/)) {
+    return 'tsx';
   }
   if (filePath.match(/\.m?jsx?$/)) {
     return 'js';
@@ -37,7 +41,6 @@ export const plugins: ParserPlugin[] = [
   'functionBind',
   'functionSent',
   'importMeta',
-  'jsx',
   'logicalAssignment',
   'nullishCoalescingOperator',
   'numericSeparator',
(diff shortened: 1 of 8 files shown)


src/parsers/helper.ts
```diff
@@ -52,11 +55,14 @@ export const plugins: ParserPlugin[] = [
...
     return {plugins: [...plugins, 'typescript']};
   }
-  const jsOptions: ParserOptions = {plugins: [...plugins, 'flow']};
+  if (fileType === 'tsx') {
+    return {plugins: [...plugins, 'typescript', 'jsx']};
+  }
+  const jsOptions: ParserOptions = {plugins: [...plugins, 'flow', 'jsx']};
   if (fileType === 'js') {
     return jsOptions;
   }
```
connectdotz: just personal opinion, maybe a switch statement is less verbose and more expressive...?

src/parsers/helper.ts
```diff
@@ -9,14 +9,18 @@
...
+ * determine if the file is a javascript (js), typescript (ts), or typescript JSX (tsx),
+ * otherwise returns undefined.
  * @param filepath
- * @returns 'js'|'ts' or undefined
+ * @returns 'js'|'ts'|'tsx' or undefined
  */
-export const supportedFileType = (filePath: string): 'ts' | 'js' | undefined => {
-  if (filePath.match(/\.tsx?$/)) {
+export const getFileType = (filePath: string): 'js' | 'ts' | 'tsx' | undefined => {
+  if (filePath.match(/\.ts$/)) {
```
connectdotz: I like your comment on the other thread about making it case-insensitive...

connectdotz (commented):
thanks for the PR, looks good to me 👍  just a few minor comments

src/parsers/helper.ts
ehaynes99: Oops, lol. Forgot about that

src/parsers/helper.ts
ehaynes99: I stuck with conditionals to be consistent with the other methods. However, on looking at it a bit more, the `getFileType` function isn't exposed, making the pair of conditionals a bit redundant. How about this version that simply collapses them?

connectdotz:
@gregoryzh do you have anything you want to add? Since #72 is similar to this PR, let's work together on one.

gregoryzh:
> @gregoryzh do you have anything you want to add? Since #72 is similar to this PR, let's work together on one.

sure,
One thing, since we are solving the Agular bracket issue for ts. @ehaynes99 can you add a test case for that?

ehaynes99:
Good catch. I had one, but must have accidentally removed it. Added.

connectdotz (approved):
nice, thanks @ehaynes99  and @gregoryzh !!

PR #79, opened 2021-09-10
allow runner to use custom shell 

connectdotz:
## motivation

We used to launch jest process with the default shell defined by Node child_process, however, there are issues like jest-community/vscode-jest#708 that would have worked in a different shell (powershell). There are also a few shell environment-related issues that could also benefit from specifying the same shell users actually use.

Also removed the `CI: true` flag in the process.env as a workaround in the past. Now, this is better served by adding an explicit jest argument `--watchAll=false` if needed. However since this could be a breaking change, we should cust a major release to be safe.

## summary
- added `shell` option in ProjectWorkspace
- remove `CI: true` flag in the runner env override. Projects can pass explicit `watchAll=false` to ensure the same effect if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 build/
 coverage/
 *.tgz
+.eslintcache
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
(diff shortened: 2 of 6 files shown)


connectdotz:
@stephtr if you have some cycle, could you try setting `shell` to powershell in windows to make sure it worked as expected? thanks.

connectdotz:
@stephtr looks like you are busy. I will merge and address issues if any.

PR #87, opened 2022-05-27
support spawn process in a login-shell for non-windows platform

connectdotz:
Add the ability to execute the jest command in a login shell, instead of the non-login/non-interactive shell, by expanding the `ProjectWorkspace.shell` config to allow a new `LoginShell` type. This feature is only supported for the non-windows platforms.

This is mainly to support the use case when the non-login/non-interactive env is not fully initialized, such as sometimes seen in vscode. 

example ProjectWorkspace.shell:
```javascript
{
  path: "/bin/bash",
  args: ["--login"],
}
```

When seeing this argument, instead of spawning the jestCommandLine directly with the "shell" option, we will spawn the login shell instead, then send jestCommandLine to the child process via stdin, followed by "exit".

Also, take the opportunity to migrate the changed files to typescript.

This is fully backward compatible, thus only bumping the minor version number.


connectdotz:
@orta looks like you are busy. I will merge this in. We can reopen it if you have any concerns later.

connectdotz:
@orta looks like you are busy. I will merge this in. We can reopen it if you have any concerns later.

PR #97, opened 2022-12-21
upgrade dependency to the latest

### LINKED ISSUE #96: Parser crashes on TypeScript inline type imports
I'm pretty sure this is the right place to create this issue, but if not let me know.

I'm trying to use vscode to debug some of my unit tests. One of my test files uses the relatively new [inline type imports](https://www.typescriptlang.org/docs/handbook/2/modules.html#inline-type-imports) feature to condense imports:

```TypeScript
import fs from 'node:fs';
import path from 'node:path';
import assert from 'node:assert';
import { EOL } from 'node:os';
import { isNativeError } from 'node:util/types';
import { declare } from '@babel/helper-plugin-utils';
import { asMockedFunction, type AnyFunction } from '@xunnamius/jest-types';
import babel from '@babel/core';
```

However, the Jest plugin for vscode crashes, giving me the following error as all IDE UI decorations disappear:

```text
/repos/babel-plugin-tester/test/plugin-tester.test.ts: failed to parse test results: SyntaxError: Unexpected token, expected "," (7:32)
```

Removing the inline type annotation fixes it. Is this expected behavior (that is: am I missing a babel plugin or something)?

Thanks for reading!

connectdotz:
Upgrade all dependencies to the latest and fix eslint/flow related errors. 

There are quite a few type erasure issues when importing typescript types in flow files that the latest flow starts to complain about. I end up adding the `$FlowIgnore` to bypass these warnings. I think we should put a consistent effort into migrating to typescript sooner than later...

fixes #96


adrienbaron:
@connectdotz Happy New Year 🥳! 

Thank you for this PR!
I'm having the same issue this fixes using the VSCode Jest extension in my repo, and looking into fixing it I realised the issue wasn't there when running the extension locally and discovered you've fixed it already 🙇‍♂️!

Is there anything I can do to help getting this and a new version of the extension using it released?

connectdotz:
@adrienbaron, Happy new year!

Sorry, still in holiday coma, should be back next week.

I was working on one more PR and plan to release a new version of this library, then vscode-jest afterward. Thanks for the offering; I will keep that in mind.

PR #99, opened 2023-01-11
add parser plugin options for decorators

### LINKED ISSUE #89: Typescript parameter decorator support
Hey everyone!

Typescript features 3 kinds of decorators - class-level, method-level and parameter-level.
Only the first two seem to be officially supported in babel 7 - https://babeljs.io/docs/en/babel-plugin-proposal-decorators
But there is a plugin available adding support for these - https://github.com/WarnerHooh/babel-plugin-parameter-decorator

So, I have just been investigating a error within a vscode extension, namely [vscode-jest-runner](https://github.com/firsttris/vscode-jest-runner) and it will fail to work correctly as soon as parameter decorators are used within jest-typescript files.
Issue here: https://github.com/firsttris/vscode-jest-runner/issues/260

Is it possible to register separate plugins to be included before executing parser?
If not, would you mind to include the plugin by default? 
Thanks!

connectdotz:
adding an optional parameter for parse function to allow users to override `decorators` plugin. For example:
```ts
parse(file, undefined, {plugins: {decorators: 'legacy'}}
parse(file, undefined, {plugins: {decorators: {decoratorsBeforeExport: false}}}
```

The same change has been made to `Snapshot` parse functions.

resolve #89


connectdotz:
@firsttris do you want to look at this PR before the merge?

connectdotz:
ok, I will merge first, a few other PR pending on this...

PR #103, opened 2023-05-21
add useDashedArgs workspace setting

mjamin:
See [this](https://github.com/jest-community/vscode-jest/pull/924#issuecomment-1292886106) for context.

### DIFF
--- a/src/Runner.js
+++ b/src/Runner.js
@@ -88,6 +88,16 @@ export default class Runner extends EventEmitter {
     if (this.options.args) {
       args.push(...this.options.args.args);
     }
+    if (this.workspace.useDashedArgs) {
+      args.forEach((arg, index) => {
+        if (!arg || !arg.startsWith('--') || arg.length === 2) {
+          return;
+        }
+
+        args[index] = arg.replace(/(\B)([A-Z])/gm, '-$2').toLowerCase();
+      });
+    }
+
     return args;
   }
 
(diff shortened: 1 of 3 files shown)


src/__tests__/runner.test.js
```diff
@@ -302,6 +302,31 @@ describe('Runner', () => {
...
+      const expected = '--test-location-in-results';
+
+      const workspace: any = {useDashedArgs: true};
+      const options = {};
+      const sut = new Runner(workspace, options);
+      sut.start(false);
+
+      const args = (createProcess: any).mock.calls[0][1];
+      expect(args[0]).toBe(expected);
+    });
```
connectdotz, line 328: let's add some test case to ensure user passed in args (options.args.args) will also be converted. Also add test cases to ensure if any of the arguments are already in dashed style, it will not be altered.

connectdotz (commented):
@mjamin Thanks for the PR, it looked good. Just some minor suggestions for new test cases.

src/Runner.js
```diff
@@ -88,6 +88,16 @@ export default class Runner extends EventEmitter {
     if (this.options.args) {
       args.push(...this.options.args.args);
     }
+    if (this.workspace.useDashedArgs) {
+      args.forEach((arg, index) => {
```
connectdotz: This code is not wrong, but usually, we use `map` for transformation operations. This might be more of a style and convention rather than correctness...

connectdotz (dismissed):
Looks good. It would be great if you could address this one minor style issue...

connectdotz (approved):
great, thanks!

rafagsiqueira:
Was this released? Still seeing this: `Error: Unknown arguments: listTests, json, watchAll` after enabling "jest.useDashedArgs": true

PR #110, opened 2023-09-30
feat: support for showing status updates as soon as they change

ffMathy:
Needed for https://github.com/jest-community/vscode-jest/pull/1075

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import type {Status} from '@jest/test-result';
 import {EventEmitter} from 'events';
 import {ChildProcess} from 'child_process';
 import {Config as JestConfig} from '@jest/types';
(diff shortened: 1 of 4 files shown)


ffMathy:
It might contain breaking changes though. Not sure if it's a problem or not - can anyone assist with deciding this?

ffMathy:
It might contain breaking changes though. Not sure if it's a problem or not - can anyone assist with deciding this?

connectdotz:
@ffMathy, thanks for the PR. 

This repository does need some cleanup. We inherited it a few years back that was mainly a javascript/flow code base. Never did fully convert to typescript, so there are lots of duplicate type def, etc. 

Clean-up is always welcome, as long as it doesn't break the logic. If it is just type-name change, then I think we are ok. But if the new/old types are not compatible, then we will need to examine them more closely and decide on a case-by-case basis... 

does it compile, and all tests passed? A good test will be to use an updated local repo with `vscode-jest` when you test the changes. It will validate if the types are compatible after changes.

ffMathy:
Interesting. Thanks for that insight.

I do unfortunately believe this is a breaking change.

I'm wondering though, are anyone other than ourselves consuming it? Do we need to retain compatibility?

I can work around it by applying the mappings instead in the vscode-jest repo, but at the cost of performance.

connectdotz:
What kind of breaking change? We can always cut a major release if needed.

Yes this repo is used by others as well, not just vscode-jest. We need to be mindful.

connectdotz:
@ffMathy To clarify, if the breaking change means to get rid of our own custom types and replace them with Jest's official types, which is a superset of the old custom type, then I think we are all right. If there is some info missing from the new type, then we should discuss those.

PR #116, opened 2024-04-11
Convert code base to Typescript

connectdotz:
## Changes
- Convert all codebase to Typescript
- Replaced most manually maintained duplicated type definitions with the original types (from jest and babel)
- Given that all the code bases are now TS, I replaced the Babel typescript compiler with `tsc`.
- Replace `babel-jest` with `ts-test` to run Jest with better type error detection.
- Removed all flow artifacts
- Upgraded eslint with prettier
- Generate `index.d.ts` from the ts files.
- Removed husky.

## Backward compatibility

This PR does not change functionality, and I try to keep the types as backward-compatible as possible. However, there are still some breaking changes (mainly types). Therefore, mark this as "32.0.0-beta.0" to start preparing for the next major release. 

## Integration Testing

I have done some preliminary tests with `vscode-jest` and it seems to work fine. Would encourage other dependent systems to try out and let me know if there are any other unexpected issues.


seanpoulter:
I'm curious why you chose to remove `husky` at the same time.

tsconfig.prod.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false, // Disable source maps for production
+    "declaration": false, // Enable declaration files generation
```
seanpoulter: The comment doesn't match the `false`.

seanpoulter:
Wow! :clap: :tada: 

I'm realizing I won't be able to review a diff this big in my spare time, especially with renaming the files and changes not being picked up
https://github.com/jest-community/jest-editor-support/pull/116/commits/3a9d429fc144f492fbe293776a425dcb1c6c3e33#diff-4fa18de8d59e8fbce3df7e88da529ad458bbf37e526d58d30b56707ba4d0d8b6

Sorry. That's as far as I'll get. :face_in_clouds:

connectdotz:
@seanpoulter, Thanks for taking a look. I know this is a tough PR to review, sorry. I gave up the piecemeal approach as it made the conversion much harder 😰...  You already helped, thank you.🙏  It's nice to know that I am not alone on this!

I figure we will bite the bullet and do this in one PR, but we won't release it to prod (thus the 32.0.0-beta.0 version), so we can have more time to test and fix when needed. Does that sound alright?

connectdotz:
> I'm curious why you chose to remove husky at the same time.

Mainly to simplify the env and reduce maintenance overhead. Given we already have lint/CI to detect errors and prettier-write to correct them, I thought husky/lint-staged didn't really add any extra value... Besides, we didn't use them in `vscode-jest` and it seems to be fine.

